PR #36, opened 2014-04-05
Combine repeated response headers by concatenating their values with a comma separator

alekstorm:
The spec isn't exactly inspiring in its clarity, but I'm pretty sure [this section](http://tools.ietf.org/html/draft-ietf-httpbis-http2-10#section-8.1.3.3) allows (or even requires?) this concatenation. Special handling for `set-cookie`/`cookie` can be added later, in the `h2-10` branch.

### DIFF
--- a/hyper/http20/response.py
+++ b/hyper/http20/response.py
@@ -62,10 +62,12 @@ class HTTP20Response(object):
         self.reason = ''
 
         # The response headers. These are determined upon creation, assigned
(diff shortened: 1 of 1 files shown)


Lukasa:
At this stage the automated merge no longer works (the codebase has moved on quite far) and @alekstorm is sufficiently busy that I don't think he can own this patch. For now, I'll close it.

PR #81, opened 2015-01-28
Adding pep8 and flakes testing

t2y:
I found there's no test for coding style. I suggest using pytest plugins as below.

* https://pypi.python.org/pypi/pytest-pep8
* https://pypi.python.org/pypi/pytest-flakes

We can set what pep8/flakes error is in `pytest.ini`. Now I set it as conservative as possible without breaking current coding style. Of course, it is also possible to be strict more.

What do you think?

### DIFF
--- a/test/test_hyper.py
+++ b/test/test_hyper.py
@@ -1919,8 +1919,8 @@ class TestResponse(object):
     def test_responses_are_context_managers(self):
         stream = DummyStream('')
 
-        with HTTP20Response([(':status', '200')], stream) as resp:  # noqa
-            pass
+        with HTTP20Response([(':status', '200')], stream) as resp:
+            assert resp
 
         assert stream.closed
 
--- a/test/test_hyper.py
+++ b/test/test_hyper.py
@@ -140,7 +140,7 @@ class TestPriorityFrame(object):
         assert f.flags == set()
         assert f.depends_on == 4
         assert f.stream_weight == 64
-        assert f.exclusive == True
+        assert f.exclusive is True
 
     def test_priority_frame_comes_on_a_stream(self):
         with pytest.raises(ValueError):
@@ -408,7 +408,7 @@ class TestHeadersFrame(object):
         assert f.data == b''
         assert f.depends_on == 4
         assert f.stream_weight == 64
-        assert f.exclusive == True
+        assert f.exclusive is True
 
     def test_headers_frame_with_priority_serializes_properly(self):
         # This test also tests that we can receive a HEADERS frame with no
@@ -868,7 +868,6 @@ class TestHPACKDecoder(object):
             (':path', '/',),
             (':authority', 'www.example.com'),
         ]
-        first_header_table = first_header_set[::-1]
         first_data = (
             b'\x82\x86\x84\x01\x8c\xf1\xe3\xc2\xe5\xf2:k\xa0\xab\x90\xf4\xff'
         )
@@ -966,7 +965,7 @@ class TestIntegerDecoding(object):
 class TestHyperConnection(object):
     def test_connections_accept_hosts_and_ports(self):
         c = HTTP20Connection(host='www.google.com', port=8080)
-        assert c.host =='www.google.com'
+        assert c.host == 'www.google.com'
         assert c.port == 8080
 
     def test_connections_can_parse_hosts_and_ports(self):
@@ -1920,7 +1919,7 @@ class TestResponse(object):
     def test_responses_are_context_managers(self):
         stream = DummyStream('')
 
-        with HTTP20Response([(':status', '200')], stream) as resp:
+        with HTTP20Response([(':status', '200')], stream) as resp:  # noqa
             pass
 
         assert stream.closed
--- a/test/test_integration.py
+++ b/test/test_integration.py
@@ -382,8 +382,9 @@ class TestHyperIntegration(SocketLevelTest):
             # We should get two packets: one connection header string, one
             # SettingsFrame. Rather than respond to the packets, send a GOAWAY
             # frame with error code 0 indicating clean shutdown.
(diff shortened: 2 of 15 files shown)


sigmavirus24:
I've never seen those extensions before. I'm uncomfortable with having so many ways to configure the tools. This can be done with a little extra work in a way that creates a new job for these tests on Travis. It will then not increase the typical test time and be a separate job in tox that will fail outside of the tests and not confuse new contributors. I can share some examples showing how to set this up.

As a side note, the change to the `MANIFEST.in` file seems entirely unrelated. I know pep8 and flake8 don't check that, so why was that change made?

test/test_integration.py
```diff
@@ -382,8 +382,8 @@ def socket_handler(listener):
             # We should get two packets: one connection header string, one
             # SettingsFrame. Rather than respond to the packets, send a GOAWAY
             # frame with error code 0 indicating clean shutdown.
-            first = sock.recv(65535)
-            second = sock.recv(65535)
+            first = sock.recv(65535)  # pragma: no flakes
```
sigmavirus24: Does pytest-flakes add this? `# prgama: no flakes` is not something included in flake8. Anyone running flake8 normally on this will still see this error.

t2y:
> It will then not increase the typical test time and be a separate job in tox that will fail outside of the tests and not confuse new contributors.

It's also good, I think. I just want to detect a trivial typo or a little mistake.

> As a side note, the change to the MANIFEST.in file seems entirely unrelated. I know pep8 and flake8 don't check that, so why was that change made?

Yes, pep8 and flake8 don't check. I just thought this style is easy to confirm the diff than previous one for the future when we would modify.

test/test_integration.py
t2y: The `# prgama: no flakes` make ignore that line. An error happens if it removed as below.

```
hyper/test/test_integration.py:385: UnusedVariable
local variable 'first' is assigned to but never used
```
See also `Ignoring certain lines in files` on https://pypi.python.org/pypi/pytest-flakes

Lukasa:
I have mixed feelings towards automatic linters. Let me have a think about this and I'll get back to you. =)

dimaqq:
My 2c:

Magic comment annotations for linters are bad.

Linting tests should always be taken with a grain of salt, tests are tests for a reason. Only core code should be lint-free.

Finally, all linters make false positives. Flake8 a few, pylint really many. These tools should ideally be combined with a service where admin can mark each warning as "yes it's lint, but let it be" so that new warnings can be easily seen.
I dunno a good service like that, does anyone have a good idea?

test/test_integration.py
```diff
@@ -417,8 +417,8 @@ def socket_handler(listener):
...
+            second = sock.recv(65535)  # pragma: no flakes
```
dimaqq: how about `assert first` and `assert second` instead of comment?

t2y:
> Magic comment annotations for linters are bad.

I agree with you. It should be removed except unavoidable matter.

> These tools should ideally be combined with a service where admin can mark each warning as "yes it's lint, but let it be" so that new warnings can be easily seen.

Though we can suggest to check by oneself with lint tool on documentation or other service (if one exists), in most case, we would forget to do so, that's me :sweat:. After some warning code was committed, we would not rather fix since lack the opportunity. It's easy to check/fix before committing some code.

test/test_integration.py
t2y: I don't know why `first` and `second` are used. I wonder whether they're explanatory variable? How to fix depends on the purpose.

Lukasa: Their use was only to do with clarity, so that I know what I'm doing in those tests.

Lukasa:
I'm still thinking about this, I haven't forgotten it. =) Just nervous about adding new tools.

test/test_integration.py
t2y: I see. In that case, as dimaqq says, assert is better.

```python
            first = sock.recv(65535)
            second = sock.recv(65535)
            assert first and second
```

test/test_hyper.py
```diff
@@ -1906,7 +1905,7 @@ def test_response_calls_stream_close(self):
     def test_responses_are_context_managers(self):
         stream = DummyStream('')
 
-        with HTTP20Response([(':status', '200')], stream) as resp:
+        with HTTP20Response([(':status', '200')], stream) as resp:  # noqa
             pass
```
t2y: maybe, it should be asserted instead of `# noqa`.

```diff
-            pass
+            assert resp
```

t2y:
@Lukasa I rebased this patch for trunk. I would lean toward adding coding-style check, but it's not strong to argue. So I don't mind rejecting this patch. What will we do now?

Lukasa:
I'm still thinking about it, but will decide today. =) Either way, the thought is hugely valuable.

t2y:
Time to decide? :clock9:

Lukasa:
Apologies for going quiet the last few days, I've been unwell.

Sadly, on balance I think I'm not going to take this. Generally speaking I think that this kind of testing is most useful when the development team is large enough to make it difficult to keep them all on the same page. Given that, at least for the moment, all code review goes through me, I think I can keep up the coding standards as-is: more importantly, I can be flexible when necessary.

Thank you very much for the thought @t2y! As always, it's appreciated! :cake: :star:

t2y:
Thank you for your decision. I follow you.

I think we should clean up the following code since I feel strange. I will send another Pull Request.

```diff
 from hyper.http20.huffman_constants import (
-    REQUEST_CODES, REQUEST_CODES_LENGTH, REQUEST_CODES, REQUEST_CODES_LENGTH
+    REQUEST_CODES, REQUEST_CODES_LENGTH
 )
```

PR #85, opened 2015-02-19
Fixed to clean up double imported constants

t2y:
Just clean up.


Lukasa:
Thanks for this! It's a legacy of a previous version of HPACK when the huffman tables were different. =)

PR #88, opened 2015-02-23
Fixed to output html on Windows avoiding print() function

### LINKED ISSUE #87: Hyper CLI script fails to load HTML on Windows
Observe:

```
C:\Users\cb2\Documents\hyper>hyper GET https://http2bin.org/html
  File "C:\Python34\Scripts\hyper-script.py", line 9, in <module>
    load_entry_point('hyper==0.2.0', 'console_scripts', 'hyper')()
  File "c:\users\cb2\documents\hyper\hyper\cli.py", line 227, in main
    print(request(args))
  File "C:\Python34\lib\encodings\cp850.py", line 19, in encode
    return codecs.charmap_encode(input,self.errors,encoding_map)[0]
UnicodeEncodeError: 'charmap' codec can't encode character '\u2014' in position
1266: character maps to <undefined>
```

t2y:
Fix #87

### DIFF
--- a/hyper/cli.py
+++ b/hyper/cli.py
@@ -6,6 +6,7 @@ hyper/cli
 Command line interface for Hyper inspired by Httpie.
 """
 import json
+import locale
 import logging
 import sys
 from argparse import ArgumentParser, RawTextHelpFormatter
@@ -20,7 +21,7 @@ from hyper.compat import urlencode, urlsplit
 
 log = logging.getLogger('hyper')
 
-FILESYSTEM_ENCODING = sys.getfilesystemencoding()
+PREFERRED_ENCODING = locale.getpreferredencoding()
 
 # Various separators used in args
 SEP_HEADERS = ':'
@@ -166,7 +167,7 @@ def set_request_data(args):
         args.url.path += '?' + urlencode(params)
 
     if body:
-        content_type = 'application/json; charset=%s' % FILESYSTEM_ENCODING
+        content_type = 'application/json; charset=%s' % PREFERRED_ENCODING
         headers.setdefault('content-type', content_type)
         args.body = json.dumps(body)
 
@@ -224,7 +225,10 @@ def request(args):
 def main(argv=None):
     args = parse_argument(argv)
     log.debug('Commandline Argument: %s', args)
-    print(request(args))
+    data = request(args)
+    sys.stdout.buffer.write(data.encode(PREFERRED_ENCODING, errors='replace'))
+    sys.stdout.buffer.write(b'\n')
+    sys.stdout.buffer.flush()
 
 
 if __name__ == '__main__':  # pragma: no cover
--- a/hyper/cli.py
+++ b/hyper/cli.py
@@ -16,7 +16,7 @@ from textwrap import dedent
 
 from hyper import HTTP20Connection
 from hyper import __version__
-from hyper.compat import urlencode, urlsplit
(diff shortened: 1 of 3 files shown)


Lukasa:
application/json does not allow charsets, so we should remove that.

Additionally, JSON must always be encoded in one of UTF-8, UTF-16 and UTF-32.

t2y:
> application/json does not allow charsets, so we should remove that.

Oh, I didn't know that. Sorry, I'll fix it.

t2y:
I noticed we also should handle the difference between py27 and py3.

Lukasa:
@t2y No reason to apologise, most people don't know it. =)

t2y:
It works well on OS X. I hope so on Windows, too ...

* python3.4
```bash
(hyper)$ hyper POST https://nghttp2.org/httpbin/post name=はいぱー version=2.0
{'args': {},
 'data': '{"name": "\\u306f\\u3044\\u3071\\u30fc", "version": "2.0"}',
 'files': {},
 'form': {},
 'headers': {'Connection': 'close',
             'Content-Length': '54',
             'Content-Type': 'application/json; charset=utf-8',
             'Host': 'httpbin',
             'Via': '2.0 nghttpx'},
 'json': {'name': 'はいぱー', 'version': '2.0'},
 'origin': '153.173.23.203',
 'url': 'https://httpbin/post'}
```

* python2.7
```bash
(hyper-py27)$ hyper POST https://nghttp2.org/httpbin/post name=はいぱー version=2.0
{u'args': {},
 u'data': u'{"version": "2.0", "name": "\\u306f\\u3044\\u3071\\u30fc"}',
 u'files': {},
 u'form': {},
 u'headers': {u'Connection': u'close',
              u'Content-Length': u'54',
              u'Content-Type': u'application/json; charset=utf-8',
              u'Host': u'httpbin',
              u'Via': u'2.0 nghttpx'},
 u'json': {u'name': u'\u306f\u3044\u3071\u30fc', u'version': u'2.0'},
 u'origin': u'153.173.23.203',
 u'url': u'https://httpbin/post'}
```

hyper/cli.py
```diff
@@ -160,13 +161,16 @@ def set_request_data(args):
...
+            if is_py2:  # pragma: no cover
+                value = value.decode(PREFERRED_ENCODING)
+            body[i.key] = value
 
     if params:
         args.url.path += '?' + urlencode(params)
 
     if body:
-        content_type = 'application/json; charset=%s' % FILESYSTEM_ENCODING
+        content_type = 'application/json; charset=utf-8'
```
Lukasa: There should be no charset at all. =) `content-type = 'application/json`.

Lukasa:
One small note, but the function is otherwise correct. =)

hyper/cli.py
sigmavirus24: For reference: https://tools.ietf.org/html/rfc7159#section-8.1 and https://tools.ietf.org/html/rfc7159#section-11. You'll note in the latter that there are no optional parameters (which is what `charset` is).

t2y: Thank you for the references. I learned a lot. Originally, I read httpie's client and I tried to tell that, but it has already been fixed. :smile: 
https://github.com/jakubroztocil/httpie/commit/6e7e2f2eeaf5d07060303735d8f183e526447f25

Lukasa:
Hurrah, perfect! :cake:

PR #89, opened 2015-02-27
Initial buffered socket readline proposal.

Lukasa:
It would be useful to be able to call `sock.readline()` for HTTP/1.1 This PR adds the function.


shazow:
> I wonder if this pull request makes you happy.

Can confirm.

PR #92, opened 2015-03-10
[WIP] HTTP/1.1

Lukasa:
This pull request tracks development of hyper's HTTP/1.1 implementation. Note that much of this is really, *really* early work, and is likely to change substantially before release. However, this is open-source, so the least I can do is do this development work in public.

Feedback always appreciated!

Here's the bare minimum of outstanding things I need to do before I want to merge this:

- [x] Connection management. We should understand `Connection: close` and friends.
- [x] Sending generators for bodies (chunked upload!)
- [x] Sending files for bodies (maybe chunked upload, who knows!)
- [x] Chunked responses
- [x] TLS
- [x] Certificate verification
- [x] Send the 'Host' header automatically!
- [x] 100% test coverage, including integration tests

Some other stuff we need, but not necessarily right away:

- Plaintext proxy support
- CONNECT proxy support
- TLS-encrypted proxy support
- Basically any goddamn thing to do with stupid proxies
- Upgrade handling
- Socket extraction (an acceptable API for releasing a socket from a HTTP response or connection)
- Expect: 100 continue processing.


Lukasa:
The outstanding chunked work can be completed by extending the `read_chunked` function to take an `amt` parameter that provides an upper bound on the size of the read. This is just like the stuff that got implemented in `urllib3`, which will turn out to work really well.

Lukasa:
Alright, I think I have completed all my tickboxes.

I want to write some release tests before I merge this, but I'm beginning to think this is nearly done. There are still some outstanding work items for cleaning up:

- [ ] We reach into hyper.http20 for a few things (exceptions etc). These should be moved to hyper.common instead.
- [ ] Some of the methods aren't quite named the same (getresponse() vs get_response()). For now, let's move to the slightly better names on the HTTP/1.1 classes.
- [ ] HTTP/1.1 responses have a few extra methods that I should really support on HTTP/2 for the purposes of the abstraction layer. The obvious one here is `read_chunked`, which should at least be easy to implement for H2.

Lukasa:
These failures are difficulties with rabbitmq's APT repository, which we only hit when we try to build nghttp2. They're intermittent, and annoying.

Lukasa:
Sadly, at this point I need to reconcile the different header handling logic between the HTTP/2 layer and the HTTP/1.1 layer.

Lukasa:
This turns out to be really painful, because the HTTP/2 layer deals with headers as unicode strings and the HTTP/1.1 layer currently does so with bytestrings. I think I can make this work by adding a unicode interface to the HTTPHeaderMap, but it's really painful. =(

Lukasa:
Ok, I think the series of commits I just put up solves that problem. It simultaneously cleans and dirties the API, which is weird.

Lukasa:
@sigmavirus24 @shazow Is there any advantage in merging-and-releasing this block of code before I put a compatibility layer in front of it?

shazow:
No dis-advantage. :)

You're infinitely more likely to never release it if you wait, vs if you release.

sigmavirus24:
Without the compatibility layer, people might start testing it and using it so you get feedback sooner. That's the only advantage I can think of.

Lukasa:
Ok, screw it, I'm going to merge this. I'll make a release that contains only this change on top of v0.2.2 (which I've just released), so people will always have a previous-stable-version to move to if for some reason this breaks them.

Deep breath people, let's merge this 71-commit monster!

PR #102, opened 2015-03-17
Stop using Twitter for documentation examples (README).

jdecuyper:
Replace *twitter.com:443* with *http2bin.org:443* on the README document.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -15,8 +15,8 @@ improved speed, lower bandwidth usage, better connection management, and more.
 
     from hyper import HTTP20Connection
 
-    conn = HTTP20Connection('twitter.com:443')
-    conn.request('GET', '/')
+    conn = HTTP20Connection('http2bin.org:443')
+    conn.request('GET', '/get')
     resp = conn.getresponse()
 
     print(resp.read())


jdecuyper:
@Lukasa 
Should all unit tests be updated too and use http2bin.org instead?

Lukasa:
@jdecuyper At some stage, I do need to move the release tests that point at Twitter to point at http2bin.org. If you feel like writing a set, I'll happily merge them. =)

Lukasa:
Also, thanks for this! :cake: :sparkles: :cookie: Feel free to open a PR adding yourself to `CONTRIBUTORS.rst`.

jdecuyper:
@Lukasa very welcome :)

- I had to use www.http2bin.org to make the demo execute properly, otherwise following error is thrown:
`hyper.ssl_compat.CertificateError: hostname u'http2bin.org' doesn't match u'www.http2bin.org'`
I guess the certificate was issued to www.htt2bin.org, documentation should be updated accordingly or should it work for both?
- To compile hyper on Ubuntu (ubuntu 3.16.0-30-generic), I had to install `libffi`, `python-dev` and `libssl`. Is it worth mentioning these as prerequesites in the docs or is it too obvious?
- What is the proper way to execute the tests?

Lukasa:
@jdecuyper Interesting. Your first two points together suggest you're using Ubuntu 2.7.X where X < 9, which means you need pyOpenSSL. Frankly, I wouldn't be surprised to find that cert verification doesn't work well with just pyOpenSSL. On Python 2.7.9 or Python 3.4 everything works smoothly and the cert validates just fine. Additionally, you no longer need any of those pre-requisites.

The proper way to execute the tests is to run tox. =) `pip install tox` will get you going, and then you can just type `tox`.

jdecuyper:
@Lukasa you are right, it is running Python 2.7.6 even though I had Python 3.4 installed, I guess installing python-dev brought me back to 2.7.6, will try to fix that, thanks!

Lukasa:
No problem!

jdecuyper:
Go it fixed, compiles and works smoothly with Python 3.4.0 on Ubuntu 14.04.1

Issue #94, opened 2015-03-12
hyper do not work on my Fedora


michal-niklas:
I have problem with running Python `hyper` library that adds support for `HTTP/2` protocol.

On my Fedora machine I installed it with `pip` and `pip3` to use it with Python 2.7.8 and Python 3.4.1. Then I copied test script that connects with twitter:

	from hyper import HTTP20Connection

	conn = HTTP20Connection('twitter.com:443')
	conn.request('GET', '/')
	resp = conn.getresponse()

	print(resp.read())

I run it using Python 2.7.8 and it ends with error:

	[mn:/plib] 1 ‡ python hyper_test.py 
	Traceback (most recent call last):
	  File "hyper_test.py", line 4, in <module>
		conn.request('GET', '/')
	  File "/usr/lib/python2.7/site-packages/hyper/http20/connection.py", line 149, in request
		self.endheaders(message_body=body, final=True, stream_id=stream_id)
	  File "/usr/lib/python2.7/site-packages/hyper/http20/connection.py", line 310, in endheaders
		self.connect()
	  File "/usr/lib/python2.7/site-packages/hyper/http20/connection.py", line 215, in connect
		self._recv_cb()
	  File "/usr/lib/python2.7/site-packages/hyper/http20/connection.py", line 580, in _recv_cb
		self._consume_single_frame()
	  File "/usr/lib/python2.7/site-packages/hyper/http20/connection.py", line 496, in _consume_single_frame
		frame, length = Frame.parse_frame_header(header)
	  File "/usr/lib/python2.7/site-packages/hyper/http20/frame.py", line 52, in parse_frame_header
		frame = FRAMES[type](stream_id)
	KeyError: 80

It ends with different error with Python 3.4.1:

	[mn:/plib] 1 ‡ python3 hyper_test.py 
	Traceback (most recent call last):
	  File "hyper_test.py", line 4, in <module>
		conn.request('GET', '/')
	  File "/usr/lib/python3.4/site-packages/hyper/http20/connection.py", line 149, in request
		self.endheaders(message_body=body, final=True, stream_id=stream_id)
	  File "/usr/lib/python3.4/site-packages/hyper/http20/connection.py", line 310, in endheaders
		self.connect()
	  File "/usr/lib/python3.4/site-packages/hyper/http20/connection.py", line 204, in connect
		sock = wrap_socket(sock, self.host)
[...]

PR #103, opened 2015-03-20
Use nghttp2 instead of twitter for release test

jdecuyper:
Update the request adapter test to hit nghttp2 instead of twitter. 
Test was failing with twitter as discussed in #94.


Lukasa:
Given that we already use nghttp2 above, can we instead use http2bin?

PR #104, opened 2015-03-22
Use http2bin instead of twitter for release test

jdecuyper:
Update the request adapter test to hit http2bin instead of twitter.
Test was failing with twitter as discussed in #94.


Lukasa:
Hurrah! :cake: :sparkles: :cookie:

Lukasa:
Thanks a bunch @jdecuyper!

Issue #96, opened 2015-03-12
Improve documentation to show examples against http2bin.org, not Twitter.


Lukasa:
Following on from #94.

PR #105, opened 2015-03-23
Use http2bin.org instead of twitter.com

jdecuyper:
Still working on #96


Lukasa:
Good spot, thanks! :cake:

jdecuyper:
Welcome!

Issue #108, opened 2015-03-27
test_hitting_http2bin_org failing


jdecuyper:
When test is run against `https://http2bin.org/status/200` following error is thrown:
```
        assert all(map(lambda r: r.status_code == 200, responses))
>       assert all(map(lambda r: r.text, responses))
E       assert all(<map object at 0x7f2ef2266f60>)
E        +  where <map object at 0x7f2ef2266f60> = map(<function TestHyperActuallyWorks.test_hitting_http2bin_org.<locals>.<lambda> at 0x7f2ef21f7268>, [<Response [200]>])
```
I have set up the test as follow:
```
s = requests.Session()
a = HTTP20Adapter()
s.mount('http://http2bin', a)
s.mount('http://www.http2bin', a)

# Here are some nice URLs.

urls = [
    'https://http2bin.org/status/200',
]
```

Lukasa:
There are a few problems here.

The first problem is that you've mounted the adapters incorrectly: they need to be mounted at `https://`, not `http://`. The other problem is that you're testing that there's data in the body, but the `/status` endpoint does not return a body. =) I recommend using a different endpoint there.

jdecuyper:
Okay, then I guess it would make more sense to run the `test_hitting_http2bin_org` test against https://http2bin.org instead of http://http2bin.org, right?

Thanks :)

Lukasa:
;) Yes, yes it would. Want to make that change?

jdecuyper:
Happy to help! Will open a PR soon.

PR #109, opened 2015-03-28
Use https instead of http when testing against htt2bin.org

jdecuyper:
As discussed in #108


Lukasa:
The test graph looks good to me! :cake: :sparkles: Thanks @jdecuyper!

Issue #112, opened 2015-04-01
HPACK logging incorrectly reports the number of consumed bytes


Lukasa:
You can see this in the debug output from running `hyper --debug https://docs.trafficserver.apache.org/en/latest/`:

```
Decoding b'\x10\x07:status\x03200@\x06server\tATS/5.3.0@\x04date\x1dWed, 01 Apr2015 07:09:02 GMT@\x0ccontent-type\ttext/html@\x0econtent-length\x0542468@\rlast-modified\x1dTue, 31 Mar 2015 20:01:35 GMT@\x04vary\x0fAccept-Encoding@\x04etag\x0f"551afd1f-a5e4"@\x08x-served\x05Nginx@\x14x-subdomain-tryfiles\x04True@\x07x-deity\nasgard-lts@\raccept-ranges\x05bytes@\x03age\x043706@\x19strict-transport-security\rmax-age=86400@\x03via2https/1.1 ATS (ApacheTrafficServer/5.3.0 [cRs f])'
Decoded 7 consuming 1 bytes.
Decoded 3 consuming 1 bytes.
Decoded (b':status', b'200'), consumed 1, indexed False
Decoded 6 consuming 1 bytes.
Decoded 9 consuming 1 bytes.
Decoded (b'server', b'ATS/5.3.0'), consumed 1, indexed True
Decoded 4 consuming 1 bytes.
Decoded 29 consuming 1 bytes.
Decoded (b'date', b'Wed, 01 Apr 2015 07:09:02 GMT'), consumed 1, indexed True
Decoded 12 consuming 1 bytes.
Decoded 9 consuming 1 bytes.
Decoded (b'content-type', b'text/html'), consumed 1, indexed True
Decoded 14 consuming 1 bytes.
Decoded 5 consuming 1 bytes.
Decoded (b'content-length', b'42468'), consumed 1, indexed True
Decoded 13 consuming 1 bytes.
Decoded 29 consuming 1 bytes.
Decoded (b'last-modified', b'Tue, 31 Mar 2015 20:01:35 GMT'), consumed 1, indexed True
Decoded 4 consuming 1 bytes.
Decoded 15 consuming 1 bytes.
Decoded (b'vary', b'Accept-Encoding'), consumed 1, indexed True
Decoded 4 consuming 1 bytes.
Decoded 15 consuming 1 bytes.
Decoded (b'etag', b'"551afd1f-a5e4"'), consumed 1, indexed True
[...]

PR #113, opened 2015-04-01
HPACK fix for consumed bytes

jdecuyper:
Fix incorrectly reported consumed bytes (as reported in #112).
Use same wording and style for all debug statements.


Lukasa:
\o/ Thanks @jdecuyper! :cake: :sparkles:

PR #115, opened 2015-04-04
Have cli support the new header map

jdecuyper:
* `request(args)`
Use `headers` instead of `getheaders()` to display the content of the header map. The output format will be improved but for now it avoids an error to be thrown.

* `get_content_type_and_charset`
Convert `content-type` header from `bytestrings` to `UTF-8` before processing it.

### DIFF
--- a/hyper/cli.py
+++ b/hyper/cli.py
@@ -200,7 +200,7 @@ def parse_argument(argv=None):
 
 def get_content_type_and_charset(response):
     charset = 'utf-8'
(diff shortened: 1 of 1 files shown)


Lukasa:
It seems some tests broke, mostly because you unconditionally look up a content-type header. Might want to change that to `.get('content-type', [])`.

jdecuyper:
Thanks @Lukasa, working on new PR!

PR #122, opened 2015-04-12
Fix cli tool/tests to be available with HTTPHeaderMap

t2y:
Now cli tool handles HTTPHeaderMap correctly. I think #115 would be fixed in this changes.


Lukasa:
\o/ Thanks @t2y!

Lukasa:
:cake: :sparkles:

PR #123, opened 2015-04-13
Breakout

Lukasa:
This change pulls out the framing layer and the HPACK layer, both of which got moved into their own repositories, and then re-vendors them.

It also removes all the tests associated with them, as they're now tested in their own libraries.

The libraries are [Lukasa/hyperframe](https://github.com/Lukasa/hyperframe) and [Lukasa/hpack](https://github.com/Lukasa/hpack).


sigmavirus24:
Breakout!

PR #124, opened 2015-04-14
Update PyOpenSSL

Lukasa:
PyOpenSSL 0.15 is out, which means we can use ALPN and NPN!


Lukasa:
Hmm, weird: PyOpenSSL is having trouble with the NPN negotiation to nghttp2. Looks like I need to get a py33 environment on my machine.

Lukasa:
Found a number of problems with the PyOpenSSL layer there.

PR #125, opened 2015-04-16
ALPN Support

### LINKED ISSUE #31: Support ALPN negotiation
I'd like to do ALPN to negotiate HTTP/2 like basically everyone else does, but I can't because the standard library doesn't support it.

PyOpenSSL might be my best bet here.

Lukasa:
Resolves #31, only 13 months after it was originally opened. In a weird twist of fate, 3.3 will be the first Python version to get ALPN support from hyper.


reaperhulk:
:+1:

PR #128, opened 2015-04-20
Update stream example in documentation

jdecuyper:
* Use `get_response` instead of `getresponse` for all 3 requests.
* Add `header` to second request example.

### DIFF
--- a/docs/source/quickstart.rst
+++ b/docs/source/quickstart.rst
@@ -100,12 +100,12 @@ For example::
 
     >>> from hyper import HTTPConnection
     >>> c = HTTPConnection('http2bin.org')
-    >>> first = c.request('GET', '/get')
-    >>> second = c.request('POST', '/post', body='key=value')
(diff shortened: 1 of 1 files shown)


docs/source/quickstart.rst
```diff
@@ -100,12 +100,12 @@ For example::
 
     >>> from hyper import HTTPConnection
     >>> c = HTTPConnection('http2bin.org')
-    >>> first = c.request('GET', '/get')
-    >>> second = c.request('POST', '/post', body='key=value')
+    >>> first = c.request('GET', '/get', body='hello')
+    >>> second = c.request('POST', '/post', headers={'key':'value'})
```
Lukasa: Can you swap these two around? It makes a bit more sense to have the body on the POST rather than the GET.

Lukasa:
Thanks for this! One small note. =) :cake:

Lukasa:
Actually, there's really no need to block this patch behind that, I'll just make that small change myself. =D

Issue #119, opened 2015-04-11
Error code registry.


Lukasa:
Right now if we get an unexpected GOAWAY we print this in the exception `hyper.http20.exceptions.ConnectionError: Encountered error 12, extra data b''.`. This is fine, but we can do better.

We should bring in the error code registry from the spec so that we can print something more like this: `hyper.http20.exceptions.ConnectionError: Error INADEQUATE_SECURITY (0xc): Negotiated TLS parameters not acceptable`.

Ideally, the exception itself should be more structured, allowing easy access to numerical error codes, textual error names, long-form error descriptions, and the extra data.

I suspect the format string is roughly going to be `"Error {s:name} ({d:errno}): {s:long_error}"`, where `long_error` is the extra data if it's available, and the description from the spec if it's not available.

jdecuyper:
@Lukasa: we could have all the 13 error codes and descriptions loaded in some static array (see [spec](https://tools.ietf.org/html/draft-ietf-httpbis-http2-17#section-11.4)) and then just use the error code to index into it.

The file could be called `error_code.py` and be kept under `hyper.http20`, any comments?

Lukasa:
Sounds good to me!

jdecuyper:
@Lukasa hola amigo!

I'm wondering if this is the way I should throw the `GoAway` error:
```
if frame.error_code != 0:
   err_name = H2_ERROR_CODE_REGISTRY[frame.error_code]['Name']
   err_no = hex(frame.error_code)
   err_description = frame.additional_data or
   H2_ERROR_CODE_REGISTRY[frame.error_code]['Description']
                      
   raise ConnectionError(
          "Encountered error %s %s: %s" %
          (err_name, err_no, err_description)
   )
```
Where `H2_ERROR_CODE_REGISTRY` is the new array containing all the error codes I have just added.

Should we make the error information available via the frame itself or have another exception for this?

I will try to add a new test that consumes a `GoAway` frame.

Lukasa:
Thanks works well. I don't think the frame itself should know about error code meanings. What you should do, however, is tolerate error codes outside the registry without exploding. =)

PR #131, opened 2015-05-04
Add HTTP2 error code registry

jdecuyper:
Make the current HTTP2 errors with name and description available to hyper.
First step towards fixing # #119.

### DIFF
--- a/hyper/http20/connection.py
+++ b/hyper/http20/connection.py
@@ -19,8 +19,8 @@ from .stream import Stream
 from .response import HTTP20Response, HTTP20Push
 from .window import FlowControlManager
 from .exceptions import ConnectionError
-from .error_code_registry import H2_ERROR_CODE_REGISTRY
 
+import hyper.http20.errors as errors
 import errno
 import logging
 import socket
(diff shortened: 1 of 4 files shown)


jdecuyper:
I have added a couple of unit tests for the error code registry and implemented the logic that reads the code registry when a `GoAway` frame is received.

hyper/http20/connection.py
```diff
@@ -375,10 +376,23 @@ def receive_frame(self, frame):
...
+                err_no = hex(frame.error_code)
+
+                if frame.error_code > 0 and \
+                   frame.error_code < len(H2_ERROR_CODE_REGISTRY):
+                    err_name = H2_ERROR_CODE_REGISTRY[frame.error_code]['Name']
+                    err_description = frame.additional_data or \
+                       H2_ERROR_CODE_REGISTRY[frame.error_code]['Description']
+                else:
+                    err_name = frame.error_code;
+                    err_description = frame.additional_data
```
Lukasa: Can we implement all this logic as a lookup function in the registry? It would be nicer for this block of code to be something more like this:

```python
try:
    name, number, description = errors.get_data(frame.error_code)
except ValueError:
    error_string = "Encountered error code %d, extra data %s" % frame.error_code, frame.additional_data
else:
    error_string = "Encountered error %s %s: %s" % (name, number, description)

raise ConnectionError(error_string)
```

It would then be ideal to include a series of tests just to confirm that

hyper/http20/connection.py
```diff
@@ -375,10 +376,23 @@ def receive_frame(self, frame):
             # shutdown and all is well. Otherwise, throw an exception.
             self.close()
 
+            # If an error occured, try to read the error description from
+            # code registry otherwise use the frame's additional data.
             if frame.error_code != 0:
+                err_no = hex(frame.error_code)
+
+                if frame.error_code > 0 and \
+                   frame.error_code < len(H2_ERROR_CODE_REGISTRY):
```
Lukasa: Try to avoid explicit line continuation characters: use parentheses instead. =)

Lukasa:
Looks good so far! Some small notes on refactoring.

Lukasa:
Also, I feel like `error_code_registry` is a really long module name: can we change it to `errors` instead?

jdecuyper:
@Lukasa thanks again for the code review :+1:

hyper/http20/connection.py
```diff
@@ -20,6 +20,7 @@
 from .window import FlowControlManager
 from .exceptions import ConnectionError
 
+import hyper.http20.errors as errors
```
Lukasa: Can this be a relative import, like the others from this module?

jdecuyper: Yes, forgot to ask you about that one. How can I do a relative import and still call `errors.get_data()`? Or should I create a class and call it as `Errors.get_data()`?

sigmavirus24: `from . import errors`

jdecuyper: Thanks @sigmavirus24!

Lukasa:
\o/ Thanks so much @jdecuyper!

:sparkles: :cake: :sparkles:

jdecuyper:
Cool!

PR #132, opened 2015-05-08
Make cli use the HTTP abstraction

jdecuyper:
Have hyper's command line utility use the `HTTPConnection` abstraction instead of `HTTP20Connection`.


Lukasa:
\o/ Quick tests on the command line suggest that this is working great, thanks @jdecuyper! :sparkles: :cake: :sparkles:

Issue #15, opened 2014-02-16
Close connections gracefully.


Lukasa:
We should send `GoAway` frames when we close a connection.

jdecuyper:
Hi @Lukasa!

Spec says: *Endpoints SHOULD send a GOAWAY frame when ending a connection, providing that circumstances permit it.*

So when shutting down gracefully we need to call `close` on all streams and then send a final `GoAway` frame on the connection control stream, something like:
```
# Close all streams
for stream in list(self.streams.values()):
   print("Closing stream with id %d" % stream.stream_id)
   stream.close()

# Send GoAway frame to the server
   self._send_cb(GoAwayFrame(0), True)
```
I think we should also close the stream for connection control with id 0, but I'm not sure yet where it is.

On a related matter, the spec also says *An endpoint that receives an unexpected stream identifier MUST respond with a connection error (Section 5.4.1) of type PROTOCOL_ERROR*. I don't think we are validating if the stream identifier actually exists or not on the client side. Code looks as follow right now:
```
self.streams[frame.stream_id].receive_frame(frame)
```
I guess we should validate first if this `stream_id` exists inside the streams collection, right?

Lukasa:
> I think we should also close the stream for connection control with id 0, but I'm not sure yet where it is.

Stream zero always exists, and there is no need to explicitly close it.

> I guess we should validate first if this stream_id exists inside the streams collection, right?

A try...except block would be better style, but yes. =)

> So when shutting down gracefully we need to call close on all streams and then send a final GoAway frame on the connection control stream

Sounds like a good idea to me!

PR #133, opened 2015-05-13
Close connections gracefully

jdecuyper:
Fix for #15

* When shutting down the connection, close all streams and send a `GoAway` frame to the endpoint.
* If an unexpected stream identifier is received then cancel stream with an error of type `PROTOCOL_ERROR`.

[Review on Reviewable](https://reviewable.io/reviews/lukasa/hyper/133)

### DIFF
--- a/hyper/http20/connection.py
+++ b/hyper/http20/connection.py
@@ -558,6 +558,7 @@ class HTTP20Connection(object):
         This builds and handles a frame.
         """
         frame.parse_body(data)
+
         log.info(
             "Received frame %s on stream %d",
             frame.__class__.__name__,
--- a/hyper/http20/connection.py
+++ b/hyper/http20/connection.py
@@ -593,9 +593,8 @@ class HTTP20Connection(object):
                 f = RstStreamFrame(frame.stream_id)
                 f.error_code = 1 # PROTOCOL_ERROR
                 self._send_cb(f)
-                error_string = ("Unexpected stream identifier %d" %
-                                  (frame.stream_id))
-                raise ProtocolError(error_string)
+                log.warning("Unexpected stream identifier %d" %
+                               (frame.stream_id))
         else:
             self.receive_frame(frame)
 
--- a/hyper/http20/connection.py
+++ b/hyper/http20/connection.py
@@ -246,7 +246,12 @@ class HTTP20Connection(object):
             stream.close()
 
         # Send GoAway frame to the server
-        self._send_cb(GoAwayFrame(0), True)
+        try:
+           self._send_cb(GoAwayFrame(0), True)
+        except:
+           log.warn(
+              "GoAway frame could not be sent: %s" % sys.exc_info()[0]
+           )
 
         if self._sock is not None:
             self._sock.close()
@@ -593,8 +598,9 @@ class HTTP20Connection(object):
                 f = RstStreamFrame(frame.stream_id)
                 f.error_code = 1 # PROTOCOL_ERROR
                 self._send_cb(f)
-                log.warning("Unexpected stream identifier %d" %
-                               (frame.stream_id))
+                log.warning(
+                    "Unexpected stream identifier %d" % (frame.stream_id)
+                )
         else:
             self.receive_frame(frame)
 
--- a/hyper/http20/connection.py
+++ b/hyper/http20/connection.py
@@ -247,11 +247,9 @@ class HTTP20Connection(object):
 
         # Send GoAway frame to the server
         try:
-           self._send_cb(GoAwayFrame(0), True)
-        except:
-           log.warn(
-              "GoAway frame could not be sent: %s" % sys.exc_info()[0]
-           )
+            self._send_cb(GoAwayFrame(0), True)
+        except Exception as e:
+            log.warn("GoAway frame could not be sent: %s" % e)
 
         if self._sock is not None:
             self._sock.close()
--- a/hyper/http20/connection.py
+++ b/hyper/http20/connection.py
@@ -240,9 +240,14 @@ class HTTP20Connection(object):
 
         :returns: Nothing.
         """
-        # Todo: we should actually clean ourselves up if possible by sending
-        # GoAway frames and closing all outstanding streams. For now this will
-        # do.
+        # Close all streams
+        for stream in list(self.streams.values()):
+            log.debug("Close stream %d" % stream.stream_id)
+            stream.close()
+
+        # Send GoAway frame to the server
+        self._send_cb(GoAwayFrame(0), True)
+
         if self._sock is not None:
             self._sock.close()
             self.__init_state()
@@ -580,7 +585,15 @@ class HTTP20Connection(object):
 
         # Work out to whom this frame should go.
         if frame.stream_id != 0:
-            self.streams[frame.stream_id].receive_frame(frame)
+            try:
+                self.streams[frame.stream_id].receive_frame(frame)
+            except KeyError:
+                # If we receive an unexpected stream identifier then we
+                # cancel the stream with an error of type PROTOCOL_ERROR
+                f = RstStreamFrame(frame.stream_id)
+                f.error_code = 1 # PROTOCOL_ERROR
+                self._send_cb(f)
+                log.warning("Unexpected stream identifier %d", frame.stream_id)
         else:
             self.receive_frame(frame)
 
--- a/hyper/http20/connection.py
+++ b/hyper/http20/connection.py
@@ -18,7 +18,7 @@ from ..packages.hpack.hpack_compat import Encoder, Decoder
 from .stream import Stream
 from .response import HTTP20Response, HTTP20Push
 from .window import FlowControlManager
-from .exceptions import ConnectionError
+from .exceptions import ConnectionError, ProtocolError
 from . import errors
 
 import errno
@@ -558,7 +558,6 @@ class HTTP20Connection(object):
         This builds and handles a frame.
         """
         frame.parse_body(data)
-
         log.info(
             "Received frame %s on stream %d",
             frame.__class__.__name__,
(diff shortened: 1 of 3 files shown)


Lukasa:
I think the tests are failing because of sad window conditions in those tests. I'll see if I can make them more reliable, but I'm afraid it'll hold up merging in the meantime. =(

jdecuyper:
@Lukasa no problem, let me know if you need any change on my side!

Lukasa:
In the short term, I've provided one small review note.

---

Review status: 0 of 2 files reviewed, 1 unresolved discussion, some commit checks failed.

---

<sup>**[hyper/http20/connection.py, line 598 \[r1\]](https://reviewable.io:443/reviews/lukasa/hyper/133#-JpNMa75kPQZF1rIJeAX)** ([raw file](https://github.com/lukasa/hyper/blob/04438f96a3d16bee52c8dc4685e30e4627fe57dd/hyper/http20/connection.py#L598)):</sup>
It's not entirely clear to me that raising a `ProtocolError` here is a good idea. It's tricky to catch and punishes the user. I think raising a WARNING Level log is the better idea.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/lukasa/hyper/133)

Lukasa:
@jdecuyper (Sorry, trying out a new review tool, and you've unfortunately become my guineapig!) /cc @sigmavirus24

Lukasa:
And apparently I forgot to actually mark the files reviewed.

---

Review status: 0 of 2 files reviewed, 1 unresolved discussion, some commit checks failed.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/lukasa/hyper/133)

sigmavirus24:
I'm a bit torn, mostly because I don't know HTTP/2 that well. The warning will be informative, but does the user want to know about this in the general case (where they likely don't have logging enabled). If the answer is "yes" it should be an exception. Otherwise, I completely defer to @Lukasa

---

Review status: all files reviewed, 1 unresolved discussion, some commit checks failed.
Reviewed files:
- hyper/http20/connection.py @ r1
- test/test_hyper.py @ r1

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/lukasa/hyper/133)

jdecuyper:
Review status: 0 of 3 files reviewed, all discussions resolved, some commit checks pending.

---

<sup>**[hyper/http20/connection.py, line 598 \[r1\]](https://reviewable.io:443/reviews/lukasa/hyper/133#-JpNMa75kPQZF1rIJeAX)** ([raw file](https://github.com/lukasa/hyper/blob/04438f96a3d16bee52c8dc4685e30e4627fe57dd/hyper/http20/connection.py#L598)):</sup>
Agreed, an unexpected stream id is not a show stopper. Issuing a warning and moving on is a better strategy.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/lukasa/hyper/133)

jdecuyper:
Review status: 0 of 3 files reviewed, 2 unresolved discussions, some commit checks pending.

---

<sup>**[hyper/http20/errors.py, line 12 \[r2\]](https://reviewable.io:443/reviews/lukasa/hyper/133#-JpOPB-J6LVgpjf-tsQW)** ([raw file](https://github.com/lukasa/hyper/blob/03d14b67cff70aebb6e77d046af4f9306faf8b4b/hyper/http20/errors.py#L12)):</sup>
Yeah, RFC is ready! Starting to work on #134

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/lukasa/hyper/133)

jdecuyper:
Still no love from the build, is is related to my specific test or is it more about some race condition between all tests?

Lukasa:
It's a race condition in the tests, I'll nail them down today. Otherwise, this is looking great.

---

Review status: all files reviewed, 3 unresolved discussions, some commit checks failed.
Reviewed files:
- hyper/http20/connection.py @ r2
- hyper/http20/errors.py @ r2
- test/test_hyper.py @ r2

---

<sup>**[hyper/http20/connection.py, line 597 \[r1\]](https://reviewable.io:443/reviews/lukasa/hyper/133#-JpNMa75kPQZF1rIJeAX)** ([raw file](https://github.com/lukasa/hyper/blob/04438f96a3d16bee52c8dc4685e30e4627fe57dd/hyper/http20/connection.py#L597)):</sup>
From a stylistic perspective I'd rather have the warning be formatted like this:

```python
log.warning(
    "Unexpected stream identifier %s" % frame.stream_id
)
```

---

<sup>**[hyper/http20/errors.py, line 12 \[r2\]](https://reviewable.io:443/reviews/lukasa/hyper/133#-JpOPB-J6LVgpjf-tsQW)** ([raw file](https://github.com/lukasa/hyper/blob/03d14b67cff70aebb6e77d046af4f9306faf8b4b/hyper/http20/errors.py#L12)):</sup>
\o/

---

<sup>**[test/test_hyper.py, line 1273 \[r2\]](https://reviewable.io:443/reviews/lukasa/hyper/133#-JpS6Xf2RbniEXfq8jBa)** ([raw file](https://github.com/lukasa/hyper/blob/03d14b67cff70aebb6e77d046af4f9306faf8b4b/test/test_hyper.py#L1273)):</sup>
Can we use a different stream ID? -1 isn't a valid one, so we can't rely on the code not rejecting this for other reasons in the future. =)

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/lukasa/hyper/133)

Lukasa:
Aha, it *is* a race condition, but it's only possible because of this patch. It's also a legitimate test failure.

We need to do two things here. First, `test_connection_context_manager` needs to expect the GoAway frame that will now be sent to it. It also needs to change its timing to force that frame to be emitted. It might be easiest for me to do that.

The other thing you *should* do is add a `try...except` block. I'll mark where in the code.

---

Review status: all files reviewed, 4 unresolved discussions, some commit checks failed.

---

<sup>**[hyper/http20/connection.py, line 249 \[r2\]](https://reviewable.io:443/reviews/lukasa/hyper/133#-JpS8aX7A35t0JZy0zHk)** ([raw file](https://github.com/lukasa/hyper/blob/03d14b67cff70aebb6e77d046af4f9306faf8b4b/hyper/http20/connection.py#L249)):</sup>
Please add a `try...except` that defends this code from basically all exceptions: we don't care if we can't emit this frame, we just want to do our best to emit it.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/lukasa/hyper/133)

jdecuyper:
Review status: 1 of 3 files reviewed, all discussions resolved, some commit checks pending.
Reviewed files:
- hyper/http20/connection.py @ r1

---

<sup>**[hyper/http20/connection.py, line 249 \[r2\]](https://reviewable.io:443/reviews/lukasa/hyper/133#-JpS8aX7A35t0JZy0zHk)** ([raw file](https://github.com/lukasa/hyper/blob/03d14b67cff70aebb6e77d046af4f9306faf8b4b/hyper/http20/connection.py#L249)):</sup>
Done.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/lukasa/hyper/133)

jdecuyper:
Review status: all files reviewed, 1 unresolved discussion, some commit checks pending.
Reviewed files:
- hyper/http20/connection.py @ r3
- test/test_hyper.py @ r3

---

<sup>**[test/test_hyper.py, line 1273 \[r2\]](https://reviewable.io:443/reviews/lukasa/hyper/133#-JpS6Xf2RbniEXfq8jBa)** ([raw file](https://github.com/lukasa/hyper/blob/03d14b67cff70aebb6e77d046af4f9306faf8b4b/test/test_hyper.py#L1273)):</sup>
2 is a more realistic stream identifer that could be received form the server.
Reviewable is doing a good job at displaying edits and comments regarding the same file accross multiple commits, pretty cool so far.   

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/lukasa/hyper/133)

Lukasa:
Alright, one tiny diff. We still have to fix the tests, but I'm happy to do that myself and point you at the diff so you can understand where the problem lies. =) It's not inherent in your code, just so you know.

---

Review status: all files reviewed, 1 unresolved discussion, some commit checks failed.
Reviewed files:
- hyper/http20/connection.py @ r3
- test/test_hyper.py @ r3

---

<sup>**[hyper/http20/connection.py, line 254 \[r3\]](https://reviewable.io:443/reviews/lukasa/hyper/133#-JpTOlofzpMLiWoijau4)** ([raw file](https://github.com/lukasa/hyper/blob/3307cba43a557693eefe974e0ae34acf40072272/hyper/http20/connection.py#L254)):</sup>
This is great, but let's tweak it a bit. =)

First, replace `except` with `except Exception`. A plain `except` catches lots of things you don't want to catch, like `KeyboardInterrupt` (caused by Ctrl+Cing a process). Best to let them pass by only catching subclasses of `Exception`.

Next, let's save ourselves some code and write the whole block:

```python
except Exception as e:
    log.warn("GoAway frame could not be sent: %s", % e)
```

This is extremely nit-picky feedback, so if you don't believe you have time to handle it just let me know and I can apply that diff myself.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/lukasa/hyper/133)

jdecuyper:
Would be 

---

Review status: 2 of 3 files reviewed, all discussions resolved, some commit checks pending.

---

<sup>**[hyper/http20/connection.py, line 254 \[r3\]](https://reviewable.io:443/reviews/lukasa/hyper/133#-JpTOlofzpMLiWoijau4)** ([raw file](https://github.com/lukasa/hyper/blob/3307cba43a557693eefe974e0ae34acf40072272/hyper/http20/connection.py#L254)):</sup>
Really appreciate the feedback, I just pushed this change.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/lukasa/hyper/133)

jdecuyper:
Would be glad to check where the problem lies, thanks!

---

Review status: 2 of 3 files reviewed, 1 unresolved discussion, some commit checks pending.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/lukasa/hyper/133)

Lukasa:
Alright, this looks great to me! :cake: :sparkles:

---

Review status: all files reviewed, all discussions resolved, some commit checks pending.
Reviewed files:
- hyper/http20/connection.py @ r4

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/lukasa/hyper/133)

Lukasa:
FYI @jdecuyper, this is the relevant commit that I believe will fix the tests. 21f611080228f4b748c515457350a11b751576a5

Issue #69, opened 2014-08-07
Enable processing for SETTINGS_MAX_FRAME_SIZE.


Lukasa:
We should be able to offer the ability to send us larger frames if users want it. This is an obvious foot gun so it needs to not be too easy to do, but it should be possible.

Suggested API:

```python
c = HTTP20Connection()
c.max_frame_size = 2**22
```

This should cause updated settings and the emission of the appropriate SETTINGS frame. Additionally, we should add validation of decoded frame lengths and reject anything that's too long.

jdecuyper:
@Lukasa

Initially the maximum frame size is `2^14-1`. This value can be changed by the endpoint (when receiving a settings frame) or by the client itself (using the `max_frime_size` property as you mention above).

When consuming a single frame (inside `connection.py`), I have added a check against the current max frame size:
```
frame_size = self._settings[SettingsFrame.SETTINGS_MAX_FRAME_SIZE]
if (length > frame_size):
    f = RstStreamFrame(frame.stream_id)
    f.error_code = 6 # FRAME_SIZE_ERROR
    self._send_cb(f)
    log.warning(
        "Frame size exceeded on stream %d (received: %d, max: %d)",
        stream_id,
        length,
        frame_size
        )
```
So far so good or any comments before continuing?  :)
I guess we should put a similar check when calling `_send_cb`?

Lukasa:
Whenever we send an `RstStreamFrame` we should delete the associated stream from our stream map. It may be worth pulling that into a helper method, as I believe we do similar things elsewhere in the code.

jdecuyper:
Yes, I'm going to add the following utility method to the connection class itself:
```
_send_rst_frame(stream_id, error_code)
   # send reset stream frame
   # remove stream from map
```
I detected already 5 potential calls to this method.

jdecuyper:
What I still need to add:
* `max_frame_size` property: using it will update the setting and cause the emission of the appropriate SETTINGS frame to the server.
* We should also check against max frame size when calling `_send_cb`, right?

Lukasa:
Let's do the `max_frame_size` property later, I don't think it's a high priority feature, and we want to handle it really carefully so that we can extend it to other settings.

As I note in my review, we definitely need to check against max frame size.

jdecuyper:
@Lukasa thanks for in depth review, will work on changes and come back.

What other feature would you consider high priority and being easy or intermediate to implement?

Lukasa:
[...]

PR #135, opened 2015-05-29
Check if length of frame is valid

jdecuyper:
Fix for #69 

* Before processing a frame, check if its length is valid using the range defined by the [spec](https://http2.github.io/http2-spec/#SETTINGS_MAX_FRAME_SIZE). 
* Allow endpoint to update the max frame size setting.

### DIFF
--- a/hyper/http20/connection.py
+++ b/hyper/http20/connection.py
@@ -12,7 +12,7 @@ from ..common.headers import HTTPHeaderMap
 from ..packages.hyperframe.frame import (
     FRAMES, DataFrame, HeadersFrame, PushPromiseFrame, RstStreamFrame,
     SettingsFrame, Frame, WindowUpdateFrame, GoAwayFrame, PingFrame,
-    BlockedFrame
+    BlockedFrame, FRAME_MAX_LEN
 )
 from ..packages.hpack.hpack_compat import Encoder, Decoder
 from .stream import Stream
@@ -123,6 +123,7 @@ class HTTP20Connection(object):
         # Values for the settings used on an HTTP/2 connection.
         self._settings = {
             SettingsFrame.INITIAL_WINDOW_SIZE: 65535,
+            SettingsFrame.SETTINGS_MAX_FRAME_SIZE: FRAME_MAX_LEN,
         }
 
         # The socket used to send data.
@@ -469,12 +470,7 @@ class HTTP20Connection(object):
         """
         Called by a stream when it would like to be 'closed'.
         """
-        if error_code is not None:
-            f = RstStreamFrame(stream_id)
-            f.error_code = error_code
-            self._send_cb(f)
-
-        del self.streams[stream_id]
+        self._send_rst_frame(stream_id, error_code)
 
     def _send_cb(self, frame, tolerate_peer_gone=False):
         """
@@ -534,6 +530,16 @@ class HTTP20Connection(object):
 
         # Parse the header. We can use the returned memoryview directly here.
         frame, length = Frame.parse_frame_header(header)
+
+        frame_size = self._settings[SettingsFrame.SETTINGS_MAX_FRAME_SIZE]
+        if (length > frame_size):
+            self._send_rst_frame(frame.stream_id, 6) # 6 = FRAME_SIZE_ERROR
+            log.warning(
+                "Frame size exceeded on stream %d (received: %d, max: %d)",
+                frame.stream_id,
+                length,
+                frame_size
+            )
 
         # Read the remaining data from the socket.
         data = self._recv_payload(length)
@@ -595,9 +601,7 @@ class HTTP20Connection(object):
                 # the ENABLE_PUSH setting is 0, but the spec leaves the client
                 # action undefined when they do it anyway. So we just refuse
                 # the stream and go about our business.
-                f = RstStreamFrame(frame.promised_stream_id)
-                f.error_code = 7 # REFUSED_STREAM
-                self._send_cb(f)
+                self._send_rst_frame(frame.promised_stream_id, 7)
 
         # Work out to whom this frame should go.
         if frame.stream_id != 0:
@@ -606,9 +610,7 @@ class HTTP20Connection(object):
             except KeyError:
                 # If we receive an unexpected stream identifier then we
                 # cancel the stream with an error of type PROTOCOL_ERROR
-                f = RstStreamFrame(frame.stream_id)
-                f.error_code = 1 # PROTOCOL_ERROR
-                self._send_cb(f)
+                self._send_rst_frame(frame.stream_id, 1)
                 log.warning(
                     "Unexpected stream identifier %d" % (frame.stream_id)
                 )
@@ -637,6 +639,21 @@ class HTTP20Connection(object):
             except ConnectionResetError:
                 break
 
+    def _send_rst_frame(self, stream_id, error_code):
+        """
+            Send reset stream frame with error code and remove stream from map.
+        """
+        if error_code is not None:
+            f = RstStreamFrame(stream_id)
+            f.error_code = error_code
+            self._send_cb(f)
+
+        try:
+            del self.streams[stream_id]
+        except KeyError as e:  # pragma: no cover
+            log.warn(
+                "Stream with id %d does not exist: %s",
+                stream_id, e)
 
     # The following two methods are the implementation of the context manager
     # protocol.
--- a/hyper/http20/connection.py
+++ b/hyper/http20/connection.py
@@ -12,7 +12,7 @@ from ..common.headers import HTTPHeaderMap
 from ..packages.hyperframe.frame import (
     FRAMES, DataFrame, HeadersFrame, PushPromiseFrame, RstStreamFrame,
     SettingsFrame, Frame, WindowUpdateFrame, GoAwayFrame, PingFrame,
-    BlockedFrame, FRAME_MAX_LEN
+    BlockedFrame, FRAME_MAX_LEN, FRAME_MAX_ALLOWED_LEN
 )
 from ..packages.hpack.hpack_compat import Encoder, Decoder
 from .stream import Stream
@@ -449,6 +449,11 @@ class HTTP20Connection(object):
             self._out_flow_control_window += delta
 
             self._settings[SettingsFrame.INITIAL_WINDOW_SIZE] = newsize
+
+        if SettingsFrame.SETTINGS_MAX_FRAME_SIZE in frame.settings:
+            new_size = frame.settings[SettingsFrame.SETTINGS_MAX_FRAME_SIZE]
+            if new_size > FRAME_MAX_LEN and new_size < FRAME_MAX_ALLOWED_LEN:
+                self._settings[SettingsFrame.SETTINGS_MAX_FRAME_SIZE] = new_size
 
     def _new_stream(self, stream_id=None, local_closed=False):
         """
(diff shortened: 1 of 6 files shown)


hyper/http20/connection.py
```diff
@@ -522,6 +523,16 @@ def _consume_single_frame(self):
...
 
+        frame_size = self._settings[SettingsFrame.SETTINGS_MAX_FRAME_SIZE]
+        if (length > frame_size):
+            self._send_rst_frame(frame.stream_id, 6) # 6 = FRAME_SIZE_ERROR
+            log.warning(
+                "Frame size exceeded on stream %d (received: %d, max: %d)",
+                frame.stream_id,
+                length,
+                frame_size
+            )
```
Lukasa: It'd be safer to log first, I think: that way the log is emitted before we explode due to any traceback.

hyper/http20/connection.py
```diff
@@ -624,6 +631,21 @@ def _recv_cb(self):
             except ConnectionResetError:
                 break
 
+    def _send_rst_frame(self, stream_id, error_code):
+        """
+            Send reset stream frame with error code and remove stream from map.
```
Lukasa: Can you remove this indentation?

hyper/http20/connection.py
```diff
@@ -624,6 +631,21 @@ def _recv_cb(self):
...
+        """
+        if error_code is not None:
+            f = RstStreamFrame(stream_id)
+            f.error_code = error_code
+            self._send_cb(f)
```
Lukasa: I don't think we need to cover the `error_code is None` case, that's surely an error.

hyper/packages/hyperframe/frame.py
```diff
@@ -10,9 +10,11 @@
 import collections
 import struct
 
-# The maximum length of a frame. Some frames have shorter maximum lengths.
+# The maximum initial lengh of a frame. Some frames have shorter maximum lengths.
 FRAME_MAX_LEN = (2 ** 14) - 1
 
+# The maximum allowed lengh of a frame.
```
Lukasa: `length`

hyper/packages/hyperframe/frame.py
```diff
@@ -10,9 +10,11 @@
 import collections
 import struct
 
-# The maximum length of a frame. Some frames have shorter maximum lengths.
+# The maximum initial lengh of a frame. Some frames have shorter maximum lengths.
```
Lukasa: `length`

hyper/http20/connection.py
```diff
@@ -436,6 +437,11 @@ def _update_settings(self, frame):
 
             self._settings[SettingsFrame.INITIAL_WINDOW_SIZE] = newsize
 
+        if SettingsFrame.SETTINGS_MAX_FRAME_SIZE in frame.settings:
+            new_size = frame.settings[SettingsFrame.SETTINGS_MAX_FRAME_SIZE]
+            if new_size > FRAME_MAX_LEN and new_size < FRAME_MAX_ALLOWED_LEN:
+                self._settings[SettingsFrame.SETTINGS_MAX_FRAME_SIZE] = new_size
+
```
Lukasa: If this value is outside the range, the spec mandates we tear the connection down with error code PROTOCOL_ERROR. Feels like we have the tools to implement that now.

test/test_hyper.py
```diff
@@ -1282,6 +1342,40 @@ def data_callback(frame):
...
+            assert error_code == 6 #FRAME_SIZE_ERROR
+
+        c = HTTP20Connection('www.google.com')
+        # Lower the maximum frame size settings in order to force the
+        # connection to send a reset frame with FRAME_SIZE_ERROR
+        c._settings[SettingsFrame.SETTINGS_MAX_FRAME_SIZE] = 10
+        c._sock = sock
+        c._send_rst_frame = send_rst_frame
+        c.request('GET', '/')
+        c._recv_cb()
```
Lukasa: This test might spuriously pass, because if `send_rst_frame` isn't called there will still be no failing asserts. Might be better to use a list that gets appended to, as I do elsewhere in this class.

Lukasa:
Brilliant stuff @jdecuyper! Great start. I've made some notes inline. =)

hyper/http20/connection.py
Lukasa: Oh, this check also isn't right. The value in `self._settings[]` is not a limit on the frame size the other side can *emit*, it's a limit on the size it can *receive*. `hyper` doesn't currently make the max frame size it can tolerate configurable (and I don't really feel like it should at this stage), so I think we just want to confirm that we tolerate MAX_FRAME_SIZE frames.

We then want to add a similar check to our own code for *sending*.

hyper/http20/connection.py
jdecuyper: Agreed, I hesitated on that one because I thought it was a bit brutal to tear down the whole connection but I guess its a safer bet.

test/test_hyper.py
jdecuyper: @Lukasa need some help here, in my test it was easy to override the max frame size setting however if use the FRAME_MAX_LEN constant (which is imported from frames), how could I easily override its value to some lower bound? Or maybe there is an easier road to take?
Thanks.

Lukasa: Is it not possible to simply manually send an oversized data frame? As in, one larger than FRAME_MAX_LEN?

jdecuyper: Thank :)

jdecuyper:
@Lukasa just to make sure I understand this right, when serializing a frame, we should be checking the length of the body against the maximum frame size. I think it would be nicer to apply this check when calling `serialize` from the send data callback. Is it okay to have serialize also return the body length?

hyper/http20/connection.py
```diff
@@ -436,6 +442,17 @@ def _update_settings(self, frame):
...
+            new_size = frame.settings[SettingsFrame.SETTINGS_MAX_FRAME_SIZE]
+            if new_size > FRAME_MAX_LEN and new_size < FRAME_MAX_ALLOWED_LEN:
+                self._settings[SettingsFrame.SETTINGS_MAX_FRAME_SIZE] = new_size
+            else:
+                log.warning(
+                    "Frame size %d is outside of allowed range",
+                    new_size)
+                # Tear the connection down with error code PROTOCOL_ERROR
+                self.close(1)
+
```
Lukasa: I think, rather than tie ourselves in knots trying to handle teardown mid-connection, we should throw exceptions whenever we have to tear a connection down ourselves. That means, still call `self.close()`, but then immediately throw an exception.

While we're here, it might be good to define an exception that we throw whenever we encounter a HTTP/2 connection error.

hyper/http20/connection.py
```diff
@@ -436,6 +442,17 @@ def _update_settings(self, frame):
 
             self._settings[SettingsFrame.INITIAL_WINDOW_SIZE] = newsize
 
+        if SettingsFrame.SETTINGS_MAX_FRAME_SIZE in frame.settings:
+            new_size = frame.settings[SettingsFrame.SETTINGS_MAX_FRAME_SIZE]
+            if new_size > FRAME_MAX_LEN and new_size < FRAME_MAX_ALLOWED_LEN:
```
Lukasa: Python is nice, so it'll let you write this:

```python
if FRAME_MAX_LEN <= new_size <= FRAME_MAX_ALLOWED_LEN:
```

test/test_hyper.py
```diff
@@ -245,6 +246,61 @@ def test_connections_increment_send_window_properly(self):
...
+        # 'Receive' the SETTINGS frame.
+        c.receive_frame(f)
+
+        # The value advertised by an endpoint MUST be between 2^14 and
+        # 2^24-1 octets. Confirm that the max frame size did not increase.
+        assert c._settings[SettingsFrame.SETTINGS_MAX_FRAME_SIZE] == FRAME_MAX_LEN
+
+        # When the setting containing the max frame size value is out of range,
+        # the spec dictates to tear down the connection.
+        assert c._sock == None
```
Lukasa: We should check that the GOAWAY frame was emitted.

Lukasa:
@jdecuyper That sounds reasonable to me =)

hyper/http20/connection.py
jdecuyper: We have already such an exception available so I have used that one.

hyper/http20/connection.py
jdecuyper: Thks

jdecuyper:
Have that function `serialize` return 2 values was breaking too many existing code so I took a different approach and expose the length of the body through `self.body_len`.

hyper/packages/hyperframe/frame.py
```diff
@@ -63,7 +65,7 @@ def parse_flags(self, flag_byte):
 
     def serialize(self):
         body = self.serialize_body()
-        body_len = len(body)
+        self.body_len = len(body)
```
Lukasa: Let's make sure we initialise this to zero in `__init__`.

jdecuyper: All set! Anything else I should cover?

hyper/http20/connection.py
```diff
@@ -482,6 +497,9 @@ def _send_cb(self, frame, tolerate_peer_gone=False):
 
         data = frame.serialize()
 
+        if frame.body_len > FRAME_MAX_LEN:  # pragma: no cover
```
Lukasa: So this is the point where we need to abort based on the set max frame size. Can you change this check to do that?

Lukasa: Sorry about having to make this note. I've had lots of trouble in the past keeping track of which setting applies to which direction in the communication, so I'm not entirely surprised that we've had to go back and forth on this a few times.

Lukasa: I also don't think there's any good reason to `#pragma: no cover` this, we can hit it pretty directly to prove the case.

Lukasa:
@jdecuyper What's the status here?

jdecuyper:
Status is: will work on fixes tomorrow if that's alright :)

Lukasa:
Sure thing. =) There's no rush, just wanted to make sure this didn't get lost! :+1:

jdecuyper:
@Lukasa 
When sending a frame, the size is checked against the maximum frame size setting (which can be updated by the server). As you asked, I have also added a test for this case.

jdecuyper:
Cool! Curious to see what the fix is about =)
In the meantime I have create the PR: https://github.com/Lukasa/hyperframe/pull/2

Lukasa:
The fix already landed, see Lukasa/hyperframe@236394736fa0b60447b7ce10965c715b0122788c.

If you rebase on top of development you should find the code you need has been vendored in from hyperframe. =D

jdecuyper:
@Lukasa the rebase worked, let me know if it looks alright to you!

Lukasa:
This is *generally* good, though the commit log has gone a bit wacky. Do you feel comfortable enough in git to fix it up, or do you want me to handle that on my end?

jdecuyper:
Sorry to bother you with this, it would be great if you could help me out here and even better if you could explain how you do it (if you have some time of course). Thanks!

Lukasa:
Of course! I'll see if I can screen record the fix.

sigmavirus24:
[Asciinema](https://asciinema.org/docs/usage) is pretty great fwiw

Lukasa:
@jdecuyper The fix was easy, as you can [see here](https://asciinema.org/a/7lrpbnf346cj0nuldz6u5em6f). All I did was take the last commit that contained actual work from you (a0316057dfa21b2656ceee0a109df0cd31d1e8ed) and merge it into the development branch. In that video you can see me scrolling down the terminal to ensure that none of your commits are there twice, which they aren't, so we know we're good.

Lukasa:
Alrighty, merged manually. Thanks @jdecuyper!

jdecuyper:
@Lukasa  many thanks for taking the time to explain :+1: :+1:

jdecuyper:
And btw thanks for the mentions in the v0.4.0 release notes =)

PR #136, opened 2015-06-02
Plaintext http20 via http11 upgrade

fredthomsen:
Still needs some cleanup and test cases need to be added and updated.

### DIFF
--- a/hyper/common/connection.py
+++ b/hyper/common/connection.py
@@ -114,22 +114,23 @@ class HTTPConnection(object):
             return self._conn.get_response()
         except HTTPUpgrade as e:
             # We upgraded via the HTTP Upgrade mechanism. We can just 
-            #go straight to the world of HTTP/2. Replace the backing object 
-            #and insert the socket into it.
+            # go straight to the world of HTTP/2. Replace the backing object 
+            # and insert the socket into it.
             assert e.negotiated == H2C_PROTOCOL
     
             self._conn = HTTP20Connection(
                 self._host, self._port, **self._h2_kwargs
             )
 
-            #stream id 1 is used by the upgrade request and response
-            self.next_stream_id += 2
             self._conn._sock = e.sock
-            
+            # stream id 1 is used by the upgrade request and response
+            # and is half-closed by the client
+            self._conn._new_stream(stream_id=1, local_closed=True)
+
             # HTTP/2 preamble must be sent after receipt of a HTTP/1.1 101
             self._conn._send_preamble()
-
-            return e.resp
+        
+            return self._conn.get_response(1)
 
     # Can anyone say 'proxy object pattern'?
     def __getattr__(self, name):
--- a/hyper/common/connection.py
+++ b/hyper/common/connection.py
@@ -7,8 +7,8 @@ Hyper's HTTP/1.1 and HTTP/2 abstraction layer.
 """
 from .exceptions import TLSUpgrade, HTTPUpgrade
 from ..http11.connection import HTTP11Connection
-from ..http20.connection import HTTP20Connection, H2C_PROTOCOL
-from ..tls import H2_NPN_PROTOCOLS
+from ..http20.connection import HTTP20Connection
+from ..tls import H2_NPN_PROTOCOLS, H2C_PROTOCOL
 
 class HTTPConnection(object):
     """
--- a/hyper/common/connection.py
+++ b/hyper/common/connection.py
@@ -5,11 +5,10 @@ hyper/common/connection
 
 Hyper's HTTP/1.1 and HTTP/2 abstraction layer.
 """
-from .exceptions import TLSUpgrade
+from .exceptions import TLSUpgrade, HTTPUpgrade
 from ..http11.connection import HTTP11Connection
-from ..http20.connection import HTTP20Connection
+from ..http20.connection import HTTP20Connection, H2C_PROTOCOL
 from ..tls import H2_NPN_PROTOCOLS
-
 
 class HTTPConnection(object):
     """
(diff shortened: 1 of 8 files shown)


Lukasa:
Fantastic @fredthomsen! I'm pretty wiped out this evening so I'll take a look tomorrow and give you some more detailed feedback.

Lukasa:
So let's start with the good news: this totally works, at least in the most simple case! I've been quite successful in getting http2bin.org to happily take the plaintext H2 upgrade, which is really awesome.

I'll start providing some code review here, but this is looking really good. @fredthomsen, if you find at any point that you don't think you want to spend the time on test cases, let me know and I'm happy to take over and clean this up and write tests.

hyper/http11/connection.py
```diff
@@ -166,6 +178,10 @@ def get_response(self):
 
         self._sock.advance_buffer(response.consumed)
 
+        if(response.status == 101 and 
+           b'upgrade' in headers['connection'] and bytes(H2C_PROTOCOL, 'utf-8') in headers['upgrade']):
```
Lukasa: The `bytes()` call here doesn't work on Python 2, so it might be better to replace it with a `.decode('utf-8')` call.

Lukasa: The `bytes()` call here doesn't work on Python 2, so it might be better to replace it with a `.decode('utf-8')` call.

hyper/http11/connection.py
```diff
@@ -130,6 +133,15 @@ def request(self, method, url, body=None, headers={}):
...
             self.connect()
 
+        #add HTTP Upgrade headers
+        headers[b'connection'] = b'Upgrade, HTTP2-Settings'
+        headers[b'upgrade'] = H2C_PROTOCOL
+        
+        #need to encode SETTINGS frame payload in Base64 and put into the HTTP-2 Settings header 
+        http2_settings = SettingsFrame(0)
+        http2_settings.settings[SettingsFrame.INITIAL_WINDOW_SIZE] = 65535
+        headers[b'HTTP2-Settings'] = base64.b64encode(http2_settings.serialize_body())
```
Lukasa: At some point we should add some heuristics about how many times we send these headers. Obviously, at a certain point we know that the upgrade is not going to happen, and we can stop adding the headers. It may be enough to only add them on the first request.

While we're here, let's try to factor this out into a separate function that handles this logic.

hyper/http11/connection.py
```diff
@@ -8,15 +8,18 @@
...
 from .response import HTTP11Response
 from ..tls import wrap_socket
 from ..common.bufsocket import BufferedSocket
-from ..common.exceptions import TLSUpgrade
+from ..common.exceptions import TLSUpgrade, HTTPUpgrade
 from ..common.headers import HTTPHeaderMap
 from ..common.util import to_bytestring
 from ..compat import bytes
 
+from ..http20.connection import H2C_PROTOCOL
```
Lukasa: It'd be better not to have an import from `http20` if at all possible. Can we pull the H2C_PROTOCOL declaration out into common code?

It might be that the best place for it is actually `tis`.

hyper/common/connection.py
```diff
@@ -93,7 +92,7 @@ def request(self, method, url, body=None, headers={}):
             # the world of HTTP/2. Replace the backing object and insert the
             # socket into it.
             assert e.negotiated in H2_NPN_PROTOCOLS
-
+
```
Lukasa: Can you remove this extra whitespace?

Lukasa:
Ok, basic code review is done.

This is totally amazing, I'm so excited to add this. I definitely owe you one @fredthomsen! :cake:

hyper/http11/connection.py
```diff
@@ -166,6 +178,10 @@ def get_response(self):
 
         self._sock.advance_buffer(response.consumed)
 
+        if(response.status == 101 and
```
Lukasa: Can you add a space to the left of this parenthesis?

hyper/http11/connection.py
fredthomsen: I am assuming you meant tls not tis.  This is where I looked first, but it seemed like the wrong place; however I can't think of a better one.

hyper/http11/connection.py
fredthomsen: Sending these all the time is wasteful, and I agree just the first request on a connection should be sufficient.

fredthomsen:
Thanks for the feedback.  I'll make these changes today.  One concern I had:
-The first stream id will never be returned for the initial request since we need to parse the response to find out if we were actually upgraded. This means that this behavior/return value differs for the request function for the `HTTPConnection` proxy object for TLS upgrades vs HTTP upgrades.

I seem to be having some issues with my tox and/or py.test setup as my test execution is hanging.  Once I get the squared up then I will address any test failures and add additional ones as needed.

hyper/http11/connection.py
Lukasa: I think it's technically the right place. 'h2c' is technically an ALPN token, even though it should never be used in ALPN, so `tls` would work fine. (And yeah, I did mean `tls`, I just didn't notice that Safari autocorrected it. :disappointed: )

Lukasa:
> The first stream id will never be returned for the initial request since we need to parse the response to find out if we were actually upgraded. This means that this behavior/return value differs for the request function for the HTTPConnection proxy object for TLS upgrades vs HTTP upgrades.

Yeah, that's annoying. Right now the API has a limitation there, in that there are no stream IDs returned if you're using HTTP/1.1. For the moment that *might* be OK.

Longer term, we may be able to extend the HTTP/1.1 API to nominally have stream IDs, but for that to always be stream ID 1, for example. Or possibly `None`.

fredthomsen:
Was looking into some of the test case failures and wondering how to handle test_hitting_http2bin_org_http11 in test_release.  This test case depended on non-secure traffic on an `HTTP11Connection` not getting upgraded.  Should there be a flag or something completely disabling the upgrade mechanism?

Lukasa:
The easiest thing to do is to simply change the host we hit. `httpbin.org` should work fine.

fredthomsen:
Ok. All the changes discussed above have been made and test cases updated to support this.  Having some issue running the integration tests, as they are just hanging.  Not sure if I have the proper threading support in my environment... Any other test cases you believe are needed for this feature?

Lukasa:
Hmm, the tests all blew up. I wonder if that's a timing problem.

Lukasa:
Those failures don't *appear* to be timing related, because they occur on my local machine as well.

fredthomsen:
Only the integration tests have an issue on my box, the rest are passing.  I'll take a look at the CI logs and try to straighten it out.

Lukasa:
Alrighty, I've worked it out. I'll do this merge manually and hopefully that'll fix up the tests.

Lukasa:
Merged! :cake: :sparkles: :cake: Thanks so much @fredthomsen! You'll find your name in CONTRIBUTORS.rst: feel free to keep contributing code, we'd love to have it!

fredthomsen:
Thanks for the help and feedback.  Were all those test case failures resolved with the change made to the `SocketLevelTest` in `server.py`?

Lukasa:
The test failures were indeed resolved, however we're short some test coverage, as mentioned in #137. If you feel like adding it let me know, otherwise I'll add those tests tomorrow.

PR #138, opened 2015-06-09
Start of http11 upgrade test

fredthomsen:
Only started on the first test. I'll wrap it up this afternoon.

### DIFF
--- a/test/server.py
+++ b/test/server.py
@@ -39,24 +39,29 @@ class SocketServerThread(threading.Thread):
                  socket_handler,
                  host='localhost',
                  ready_event=None,
-                 h2=True):
+                 h2=True,
+                 secure=True):
         threading.Thread.__init__(self)
 
         self.socket_handler = socket_handler
         self.host = host
+        self.secure = secure
         self.ready_event = ready_event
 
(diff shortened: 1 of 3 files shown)


fredthomsen:
The upgrade test fits nicely into `test/test_integration_http11.py`, but where is a good place for the standard plaintext HTTP2.0 test?

Lukasa:
Given the changes you're making, I'd say it's going to be in test_integration.py (which is implicitly HTTP/2 focused).

fredthomsen:
In that case should all those tests be run a second time with plaintext http or just treat the insecure case as a one-off test?

Lukasa:
Goood question.

Ideally I'd like us to be able to do all the tests twice, once secure and once not. A py.test fixture should give us the power to do that. For now, though, let's do a one-off and take a work item to investigate having the tests do both secure and insecure work.

fredthomsen:
Go ahead and look this over.  I think it's ready.

Lukasa:
Alrighty this looks good to me. Thanks @fredthomsen!

I don't think there are any other dependencies on the integration tests, so they really shouldn't hang. I assume you installed all of test_requires.txt?

PR #155, opened 2015-08-05
Use scheme to determine TLS use from CLI

### LINKED ISSUE #151: TLS is not used when connecting via hyper CLI to port other then 443
HttpConnection does not use TLS by default (and I haven't found a way to force it through CLI): https://github.com/Lukasa/hyper/blob/development/hyper/common/connection.py#L29-L31
I expected hyper to use TLS when connecting via CLI: ```hyper GET https://myserver:3000/```, since I've mentioned ```https```.

Also, I think that might be a security issue because hyper is mostly used for HTTP/2 communication and it is expected for all such communication to be secured. Maybe its better to have ```security=True``` by default?

Lukasa:
This resolves #151.

@direvius, you may want to check this out to confirm it resolves your first issue.

### DIFF
--- a/hyper/cli.py
+++ b/hyper/cli.py
@@ -123,6 +123,7 @@ def set_url_info(args):
             self.port = 443
             self.query = None
             self.scheme = 'https'
+            self.secure = False
 
     info = UrlInfo()
     _result = urlsplit(args._url)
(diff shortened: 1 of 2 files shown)


direvius:
Yep, I think it works, thank you!

Lukasa:
Ok, bullied Travis into passing that test. Thanks for the report @direvius!

PR #156, opened 2015-08-05
Stop being clever with bytes in CLI tool.

### LINKED ISSUE #154: Hyper CLI tries to decode 'application/octet-stream' using utf-8 codec
Response headers:
```
HTTPHeaderMap([(b'content-type', b'application/octet-stream') ... etc ...
```
Exception:
```
  File "/usr/local/bin/hyper", line 9, in <module>
    load_entry_point('hyper==0.4.0', 'console_scripts', 'hyper')()
  File "/Users/direvius/git/hyper/hyper/cli.py", line 231, in main
    data = request(args)
  File "/Users/direvius/git/hyper/hyper/cli.py", line 222, in request
    data = response.read().decode(charset)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 0: invalid start byte
```

Lukasa:
Resolves #154.

@direvius, you're up again: care to check that this solves your problem?

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -16,6 +16,8 @@ dev
   a stream.
 - Prevent hyper from emitting lots of RST_STREAM frames.
 - Hyper CLI tool now correctly uses TLS for any ``https``-schemed URL.
+- Hyper CLI tool no longer attempts to decode bytes, instead writing them
+  straight to the terminal.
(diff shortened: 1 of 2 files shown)


direvius:
Yep, it works! Thank you!

PR #157, opened 2015-08-18
SocketServerThread times out after 1s

ijl:
If there was a problem opening the socket, tests would otherwise hang indefinitely.


Lukasa:
Hmm, the PyPy tests appear to be busted. I'm honestly not sure what happened there, I'll see if I can fix them up.

ijl:
Closing to not keep the repository around.

PR #163, opened 2015-08-28
Work around cryptography support in old PyPy

### LINKED ISSUE #161: PyPy builds are broken
See [this build](https://travis-ci.org/Lukasa/hyper/builds/77454850) for more.

Lukasa:
Fixes #161.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,11 +29,23 @@ if sys.argv[-1] == 'publish':
 py_version = sys.version_info[:2]
 py_long_version = sys.version_info[:3]
 
+try:
+    pypy_version = sys.pypy_version_info[:2]
+except AttributeError:
+    pypy_version = None
+
+
 def resolve_install_requires():
-    if py_version == (3,3):
+    if py_version == (3, 3):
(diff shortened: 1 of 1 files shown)


sigmavirus24:
Word of caution: This will not work for wheels.

Lukasa:
Yeah. =( Frankly, the `setup.py` in this module is being a little bit over-clever, but I'm not really confident that we can do anything else.

sigmavirus24:
Specifically what I meant is the following:

If you build a wheel on CPython 2.7, the requirements will statically say "cryptography 1.0 is okay" and someone installing that wheel on PyPy < 2.6.0 will get that requirement and it will cause them to install 1.0 and for things to be broken.

I would be careful about publishing wheels at this point because of this.

Lukasa:
Yeah, I got that.

What I mean is that the `setup.py` has a number of flaws exactly like that, because I publish universal wheels. This is altogether somewhat problematic.

sigmavirus24:
Yeah, I would say you should stop publishing universal wheels because of this =(. pip 7.x.y will build wheels and cache them for you so they'll be 100% appropriate for each user.

PR #164, opened 2015-08-31
Add proxy support

fredthomsen:
Just making sure I am on the right track. Test cases are coming.


hyper/http11/connection.py
```diff
@@ -93,11 +105,16 @@ def connect(self):
         :returns: Nothing.
         """
         if self._sock is None:
-            sock = socket.create_connection((self.host, self.port), 5)
+            if not self.proxy_host:
+                host = self.host
+            else:
+                port = self.proxy_host
```
Lukasa: This is not quite right: you assign only host in the top block and only port in the bottom block. Need to do both in both. =)

hyper/http11/connection.py
```diff
@@ -93,11 +105,16 @@ def connect(self):
...
+                
+            sock = socket.create_connection((host, port), 5)
             proto = None
 
             if self.secure:
-                sock, proto = wrap_socket(sock, self.host, self.ssl_context)
+                sock, proto = wrap_socket(sock, host, self.ssl_context)
```
Lukasa: This is not necessarily quite right.

The 'host' part of `wrap_socket` is used to validate TLS connections, which means it should always be the remote host, not the proxy host. Regardless, I don't think we'll support TLS connections directly to proxies without the CONNECT verb just yet. For that reason you might just want to temporarily turn this function off (`assert not self.proxy_host` in this block).

hyper/http20/connection.py
```diff
@@ -222,10 +234,15 @@ def connect(self):
...
+            if not self.proxy_host:
+                host = self.host
+            else:
+                port = self.proxy_host
+
+            sock = socket.create_connection((host, port), 5)
 
             if self.secure:
-                sock, proto = wrap_socket(sock, self.host, self.ssl_context)
+                sock, proto = wrap_socket(sock, host, self.ssl_context)
```
Lukasa: These deltas have the same notes as the HTTP/1.1 version.

test/test_abstraction.py
```diff
@@ -8,19 +8,20 @@ class TestHTTPConnection(object):
...
             'ssl_context': False,
+            'proxies': False,
             'other_kwarg': True,
         }
 
     def test_h2_kwargs(self):
         c = HTTPConnection(
             'test', 443, secure=False, window_manager=True, enable_push=True,
-            ssl_context=True, other_kwarg=True
+            ssl_context=True, proxies=False, other_kwarg=True
```
Lukasa: These kwargs don't match the one you've actually used: `proxy`.

Lukasa:
Yup, this is definitely on the right track! I've left some notes inline, but I'm excited to see this work progress!

hyper/http11/connection.py
fredthomsen: I did overlook this and forget about connect.  Will go ahead and use assert like you suggested.

fredthomsen:
I am fixing this test case and adding one for http2, but before I go further do you agree with this test as it is structured?

Lukasa:
I think the test is overcomplicated. I don't think we need to run something that amounts to an actual proxy: we just need to run something that knows it is pretending to be a proxy. What we really want to test is that, when we set up proxy values, we connect to the right place and then that we format the data correctly. This doesn't require a (fragile) socket proxy, just requires the "server" side of the connection to look for some slightly different stuff.

fredthomsen:
Replaced it with a simpler test case. Think everything should be ready now. Take a look.

hyper/http11/connection.py
```diff
@@ -75,6 +79,19 @@ def __init__(self, host, port=None, secure=None, ssl_context=None,
...
+            if proxy_port is None:
+                try:
+                    self.proxy_host, self.proxy_port = proxy_host.split(':')
+                    self.proxy_port = int(self.proxy_port)
+                except ValueError:
+                    self.proxy_host, self.proxy_port = proxy_host, 8080
+            else:
+                self.proxy_host, self.proxy_port = proxy_host, proxy_port
+        else:
+            self.proxy_host = None
```
Lukasa: Can we `None` out `self.proxy_port` here as well? I hate having attributes that only *sometimes* exist.

hyper/http11/connection.py
```diff
@@ -75,6 +79,19 @@ def __init__(self, host, port=None, secure=None, ssl_context=None,
         self.ssl_context = ssl_context
         self._sock = None
 
+        # Setup proxy details if applicable.
+        if proxy_host:
+            if proxy_port is None:
+                try:
+                    self.proxy_host, self.proxy_port = proxy_host.split(':')
```
Lukasa: Can we pull this line out of the `try...except`? It isn't the one raising the `ValueError`, and tightly scoped `try` blocks are good.

hyper/http11/connection.py
```diff
@@ -75,6 +79,19 @@ def __init__(self, host, port=None, secure=None, ssl_context=None,
...
+                    self.proxy_port = int(self.proxy_port)
```
Lukasa: Can we pull this line out of the `try...except`, and into an `else`? It isn't the one raising the `ValueError` we're solving here, and tightly scoped try blocks are good.

Lukasa:
Alright @fredthomsen, this is looking really great! A few small comments inline, but we're basically there! Please also add yourself to the `CONTRIBUTORS` file. =)

fredthomsen:
Made these changes; however, looks like some strange build issues relating to DNS and SSL hit the CI process.  Force a re-run?

Lukasa:
Runs clean @fredthomsen! Let me do one last review.

hyper/http11/connection.py
```diff
@@ -75,6 +79,21 @@ def __init__(self, host, port=None, secure=None, ssl_context=None,
         self.ssl_context = ssl_context
         self._sock = None
 
+        # Setup proxy details if applicable.
+        if proxy_host:
+            if proxy_port is None:
+                try:
+                    self.proxy_host, self.proxy_port = proxy_host.split(':')
```
Lukasa: This logic could go wrong with IPv6 addresses. I think we might need to pull this out into a helper function, because the logic is a bit tricky, but we can run on some basic heuristics: if the host ends with square bracket it doesn't contain a port. Otherwise, change the split into `rsplit` only allowing one split. That should work in more cases.

test/server.py
```diff
@@ -106,15 +107,23 @@ def _start_server(self, socket_handler):
...
         self.port = self.server_thread.port
         self.secure = self.server_thread.secure
 
     def get_connection(self):
         if self.h2:
-            return HTTP20Connection(self.host, self.port, self.secure)
+            if not self.proxy:
+                return HTTP20Connection(self.host, self.port, self.secure)
+            else:
+                return HTTP20Connection('http2bin.org', secure=self.secure, proxy_host=self.host, proxy_port=self.port)
```
Lukasa: This line is overlong, can you trim it down in the same style as the rest of the codebase?

Lukasa:
Sorry @fredthomsen, a few really small notes. Shout if any point you no longer want to keep polishing this and I can do the rest of it.

hyper/http11/connection.py
fredthomsen: Agreed this won't work with IPv6 addresses; however, the rest of the codebase or at least the http11 and http20 constructors have similar logic when just parsing the host and port.  Should these all be updated as part of this effort and test cases added as well?

test/server.py
fredthomsen: Sure. You are sticking to 80 width it appears right?  I trimmed a few other stragglers I saw that I had messed up as well.

fredthomsen:
No worries.  I am very particular as well.  Don't mind doing it at all.

hyper/http11/connection.py
Lukasa: Hmm, if everything else is wrong too then I'd skip it for now. Let's open a GitHub issue to track it instead: see #170.

Lukasa:
Alright, beautiful, I'm happy with this. Thanks so much @fredthomsen!

PR #172, opened 2015-10-16
Validate ipv6 support

fredthomsen:


Lukasa:
So far you're encountering test failures, but I'm keeping an eye on the progress. Thanks for doing this work @fredthomsen!

fredthomsen:
Had some trouble getting some things to split proper and ended up with a regex.  Just wanted to get your thoughts,

Lukasa:
Hmmm, I have pretty mixed feelings about using a regex here. @sigmavirus24, do you have anything useful for this floating around?

sigmavirus24:
I have a very complicated regular expression that is built up progressively in rfc3986 and is based on the ABNF from RFC 3986
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

sigmavirus24:
https://github.com/sigmavirus24/rfc3986/blob/master/rfc3986/misc.py#L83..L114 is what I was talking about

Lukasa:
So, should we either vendor or require that package?

sigmavirus24:
Sorry I never answered you @Lukasa.

You can do either. `rfc3986` is vendorable. It has stricter support for parsing URIs according to RFC 3986 than the `urlparse` library does. You can actually replace that with `rfc3986` if you want.

Lukasa:
It feels like we should do that then. @fredthomsen, does that sound like something you want to take on?

fredthomsen:
Yes. I'll handle this.  I'll grab some help from you if needed.

On Fri, Oct 23, 2015 at 12:22 PM, Cory Benfield <notifications@github.com>
wrote:

> It feels like we should do that then. @fredthomsen
> <https://github.com/fredthomsen>, does that sound like something you want
> to take on?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Lukasa/hyper/pull/172#issuecomment-150624369>.
>



-- 
-Fred Thomsen

sigmavirus24:
:+1: @fredthomsen

fredthomsen:
When making something vendorable, what is the appropriate thing to do in terms of license files, author files, etc?  I see other vendorable packages we include have no such files.

Lukasa:
Generally speaking, when vendoring packages you should pull specific releases, include their copyright files in with them, and also add them to a NOTICES file at the root of this repo.

In the case of the other projects, I didn't do that because their license and copyright statements are exactly the same as for this project.

fredthomsen:
Seems like this library needs a full URI instead of just an IP to do it's job

`
>>> uri = uri_reference('1.2.3.4')
>>> uri.is_valid()
True
>>> uri.host
>>> uri.port
>>> uri.path
u'1.2.3.4
`

and IPv6 addresses can't be parsed without brackets which is the same as the point above.

`hyper.packages.rfc3986.exceptions.InvalidPort: The port ("dcba::1234") is not valid.`

I can work around the first problem by prepending schemes; however, the second is that IPv6 addresses will require brackets even when no port is present which seems like a strange restriction to enforce,

Lukasa:
I think it's ok to require that the IP addresses be surrounded by square brackets. That's an easy enough check to do before passing the IP to the library.

sigmavirus24:
You can also create a `URIReference` directly if you only want to validate the host. Just note that the parameter in the reference you're looking for is called `authority` (which encompasses the `userinfo`, `host`, and `port`).

fredthomsen:
@sigmavirus24 thanks for the help, I took that approach.  @Lukasa, I updated the test cases accordingly to reflect what you said.  Pushed the changes, but I am having some issues with pypy failures.

hyper/cli.py
```diff
@@ -113,8 +114,8 @@ def make_troubleshooting_argument(parser):
 def set_url_info(args):
     def split_host_and_port(hostname):
         if ':' in hostname:
-            host, port = hostname.split(':')
-            return host, int(port)
+            host, port = to_host_port_tuple(hostname)
+            return host, port
```
Lukasa: We may as well condense this to a single logical line now.

hyper/common/util.py
```diff
@@ -25,3 +27,22 @@ def to_bytestring_tuple(*x):
...
+                       query=None,
+                       fragment=None)
+
+    host = uri.host.strip('[]')
+    if not uri.port:
+        port = default_port
+    else:
+        port = int(uri.port)
+
+    return ((host, port))
```
Lukasa: We don't need the double parens here.

hyper/common/util.py
```diff
@@ -25,3 +27,22 @@ def to_bytestring_tuple(*x):
...
+def to_host_port_tuple(host_port_str, default_port=80):
+    """
+    Converts the given string containing a host and possibly a port
+    to a tuple.
+    """
+    uri = URIReference(scheme=None,
+                       authority=host_port_str,
+                       path=None,
+                       query=None,
+                       fragment=None)
```
Lukasa: The style we tend to use for this is:

```python
uri = URIReference(
    scheme=None,
    authority=host_port_str,
    path=None,
    query=None,
    fragment=None
)
```

hyper/http11/connection.py
```diff
@@ -56,11 +56,7 @@ class HTTP11Connection(object):
     def __init__(self, host, port=None, secure=None, ssl_context=None, 
                  proxy_host=None, proxy_port=None, **kwargs):
         if port is None:
-            try:
-                self.host, self.port = host.split(':')
-                self.port = int(self.port)
-            except ValueError:
-                self.host, self.port = host, 80
+            self.host, self.port = to_host_port_tuple(host)
```
Lukasa: Let's be explicit about the default port here too.

sigmavirus24:
@fredthomsen the pypy failures are because you need to add a line to [the list of packages](https://github.com/Lukasa/hyper/blob/development/setup.py#L75) otherwise it won't be installed properly.

fredthomsen:
Thanks @sigmavirus24.  Ok think we're good now.

sigmavirus24:
:cake:

Lukasa:
Ok, I officially love this and think it's beautiful. Thanks for the fantastic library @sigmavirus24, and thanks for the hard work @fredthomsen. You two are magical people. :sparkles: :cake: :sparkles:

PR #174, opened 2015-10-28
Include tox.ini in sdist

koobs:
Allows downstream consumers and OS packagers & porters to use tox for testing since py.test by itself doesn't work without hyper being installed in site-packages. Fixing the latter would be much preferable than having to run tox.

Supporting the setup.py test command and having tests *just work* via that method (read tests_require, test_suite) would also be great but is out of scope for this change:

[Pytest: Integration with setuptools test commands](https://pytest.org/latest/goodpractises.html#integration-with-setuptools-test-commands)


Lukasa:
Well there's certainly no reason not to just immediately merge this.

PR #179, opened 2015-12-07
Allow non-dictionary header arguments in request method.

irvind:
It's just a tiny improvement for HTTP11Connection class. Now you can pass any iterable as a header argument.

### DIFF
--- a/CONTRIBUTORS.rst
+++ b/CONTRIBUTORS.rst
@@ -38,3 +38,7 @@ In chronological order:
   - Added support for upgrade of plaintext HTTP/1.1 to plaintext HTTP/2.
   - Added proxy support.
   - Improved IPv6 support.
+
+- Eugene Obukhov (@irvind)
+
+  - General code improvements.
--- a/hyper/http11/connection.py
+++ b/hyper/http11/connection.py
@@ -10,6 +10,8 @@ import os
 import socket
 import base64
 
+from collections import Iterable
+
 from .response import HTTP11Response
 from ..tls import wrap_socket, H2C_PROTOCOL
 from ..common.bufsocket import BufferedSocket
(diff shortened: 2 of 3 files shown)


sigmavirus24:
So in general I'm +1 on this idea, but I'm also -0.5 on the implementation. If we want to support iterables that are orderable, we should preserve that order on the backend if possible. Users will expect that if they pass headers like

```
[('Cookie', 'some_cookie=value'),
 ('Cookie', 'some_other_cookie=some_other_value')]
```

Then they would be sent in that order. Further, the above list would remove the ability to send both of those. I haven't finished reading 7540 though, so I don't even know if that's actually a concern for h2.

Lukasa:
Thanks for this @irvind! :sparkles: :cake: :sparkles:

So, I agree with @sigmavirus24 here as well. I think the general approach here is good, but we should try to do it right. Happily, the "order in the backend" preservation logic is easily achieved by casting the `Iterable` to `HTTPHeaderMap` instead of to `dict`. That way, `hyper` gets to use a consistent object through its backend, and users can pass us whatever they want.

I think the way to do this is to extend your unit test to use more than one header. If you did this in the current change, that test should start intermittently failing because of hash randomization. Then, switching the cast to `dict` to an initialization of a `HTTPHeaderMap` should cause that test to start consistently passing.

BTW, @irvind, you should add yourself to the CONTRIBUTORS file!

irvind:
@sigmavirus24 @Lukasa Thanks for your remarks! I've tried to fix those issues in the new commit.

Lukasa:
Fab @irvind, that's much better. Let's add one final thing to that test, just to clarify that it really does work as advertised: let's change the list of sent headers to:

```python
('User-Agent', 'hyper'), 
('Custom-field', 'test'),
('Custom-field2', 'test'),
('Custom-field', 'test2'),
```

And validate that those two `Custom-field` headers really do come out as separate lines, in that order. Then I'll happily merge. =)

irvind:
@Lukasa All right, here it is.

Lukasa:
Beautiful, thanks so much @irvind! :sparkles: :cake: :sparkles:

sigmavirus24:
Woot! Good job @irvind! :+1:

PR #180, opened 2015-12-10
Add simple concurrency support to HTTP20Connection

tbetbetbe:
- adds simple scheme for allowing HTTP20Connections to be used
concurrently by adding a couple of RLocks.

- connection and stream creation is synchronized
- reads and writes of frames on the connection and its streams are
  synchronized
- connection#request is synchronized

The current tox tests pass, but no new unit tests are added.
However a threading test is added to test_release.py that passes, e.g

$ py.test -s \
test_release.py::TestHyperActuallyWorks::test_threaded_abusing_nghttp2_org

### DIFF
--- a/hyper/http20/connection.py
+++ b/hyper/http20/connection.py
@@ -376,12 +376,12 @@ class HTTP20Connection(object):
                 stream.close(error_code)
 
             # Send GoAway frame to the server
-            if self._sock is not None:
-                try:
-                    self._send_cb(GoAwayFrame(0), True)
-                except Exception as e:  # pragma: no cover
-                    log.warn("GoAway frame could not be sent: %s" % e)
+            try:
+                self._send_cb(GoAwayFrame(0), True)
+            except Exception as e:  # pragma: no cover
+                log.warn("GoAway frame could not be sent: %s" % e)
 
+            if self._sock is not None:
                 self._sock.close()
                 self.__init_state()
 
--- a/hyper/http20/connection.py
+++ b/hyper/http20/connection.py
@@ -128,6 +128,12 @@ class HTTP20Connection(object):
         # be synchronized, it uses _send_cb internally (which is serialized);
         # it's other activity (safe deletion of the stream from self.streams)
         # does not require synchronization.
+        #
+        # _read_lock may be acquired when already holding the _write_lock,
+        # when they both held it is always by acquiring _write_lock first
+        #
+        # either _read_lock or _write_lock may be acquired whilst holding _lock
+        # which should always be acquired before either of the other two.
         self._lock = threading.RLock()
         self._write_lock = threading.RLock()
         self._read_lock = threading.RLock()
@@ -836,8 +842,7 @@ class HTTP20Connection(object):
         """
         # Concurrency
         #
-        # Synchronizes reading data, using the same rentrant lock as when data
-        # is written. i.e, reads and writes to the connection are serialized.
+        # Synchronizes reading data
         #
         # i/o occurs while the lock is held; waiting threads will see a delay.
         with self._read_lock:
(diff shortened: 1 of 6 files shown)


hyper/http20/connection.py
```diff
@@ -275,23 +346,36 @@ def close(self, error_code=None):
...
+        #
+        # i/o occurs while the lock is held; waiting threads will see a delay.
+        with self._lock:
+            # Close all streams
+            for stream in list(self.streams.values()):
+                log.debug("Close stream %d" % stream.stream_id)
+                stream.close(error_code)
+
+            # Send GoAway frame to the server
+            if self._sock is not None:
```
Lukasa: There's a behavioural change here: what's the reasoning for it?

hyper/http20/connection.py
```diff
@@ -544,23 +652,12 @@ def _send_cb(self, frame, tolerate_peer_gone=False):
...
-            # If we don't have room in the flow control window, we need to look
-            # for a Window Update frame.
-            while self._out_flow_control_window < len(frame.data):
-                self._recv_cb()
-
-            self._out_flow_control_window -= len(frame.data)
-
         data = frame.serialize()
-
         max_frame_size = self._settings[SettingsFrame.SETTINGS_MAX_FRAME_SIZE]
```
Lukasa: So anything that comes off of settings can be changed under our feet. I suspect this means we'll need to have an RWLock around settings access that ensures that while we're using a settings value it doesn't change.

hyper/http20/connection.py
```diff
@@ -569,12 +666,28 @@ def _send_cb(self, frame, tolerate_peer_gone=False):
...
-                e.errno not in (errno.EPIPE, errno.ECONNRESET)):
-                raise
+        # Concurrency
+        #
+        # Synchronizes data writes.  This also ensures that only the current
+        # thread waits for window updates when the flow control window is
+        # blocked.
+        #
+        # i/o occurs while the lock is held; waiting threads will see a delay.
+        with self._write_lock:
```
Lukasa: Do we not also require the read_lock here?

Lukasa: Oh, nevermind, I see that `recv_cb` takes it itself.

hyper/http20/connection.py
```diff
@@ -98,6 +99,39 @@ def __init__(self, host, port=None, secure=None, window_manager=None, enable_pus
...
+        # I.e, sends/receve on the connection and its streams are serialized
+        # and separately across the threads accessing the connection.  This is
+        # a simple way of providing thread-safety.
+        #
+        # _write_lock and _read_lock synchronize all interactions between
+        # streams and the connnection.  There is a third i/o callback,
+        # _close_stream, passed to a stream's constructor.  It does not need to
+        # be synchronized, it uses _send_cb internally (which is serialized);
+        # it's other activity (safe deletion of the stream from self.streams)
+        # does not require synchronization.
```
Lukasa: It might be worth updating the comment here to talk about the correct order to obtain these locks in the event that more than one is required. This is particularly true for hyper, which gets a bit carried away about calling back into itself recursively. ;)

Lukasa:
Thanks for this @tbetbetbe, this is a really good start! I've just pushed up the `concurrency` branch: feel free to target this PR against it.

We need to be really careful to guard against the risk of deadlock, where the big problem is the order in which we take locks, and ensuring that some of the recursive logic used by hyper doesn't bump into any of those conditions. Doing this is going to be really tricky, I think: fundamentally, we may have to analyse possible control flow paths by eye and confirm that they have no risk of deadlocking.

hyper/http20/connection.py
tbetbetbe: Add that below

hyper/http20/connection.py
tbetbetbe: It was unnecessary, I've removed it.

Lukasa:
Closing for #189.

Issue #61, opened 2014-06-15
Use memoryviews in HPACK


Lukasa:
HPACK should use memoryview objects when decoding if at all possible to avoid copying memory all over the place.

Lukasa:
Done! Thanks @irvind!

PR #181, opened 2015-12-12
Use memoryview when decoding HPACK headers

irvind:
In this pull request I was trying to resolve the issue #61. Tests passed successfully for py3.4 and py2.7 on my machine, but unfortunately, I don't know if it's gonna work for PyPy.


Lukasa:
Hey @irvind! Unfortunately, `hpack` is actually a separate module that I vendor in. Can you make this change against the module [over here](https://github.com/python-hyper/hpack)?

irvind:
@Lukasa sure, I'll redo this change.

PR #184, opened 2015-12-22
Fixed typo in SSLContext

benlast:
In `hyper.ssl_compat.SSLContext#load_cert_chain`, the reference to `self._ctx.set_password_cb` should be `self._ctx.set_passwd_cb`.
Added `test/test_hyper_SSLContext.py`, with test client cert and key, to reproduce the issue, fixed the typo, verified that all tests pass.


Lukasa:
So, the PyPy tests are failing: this looks specifically like the problem is that you're unconditionally importing `ssl_compat` to get hold of a SSLContext without having PyOpenSSL installed (as the standard tests do not).

What we probably want to do is to skip the tests if the import of `ssl_compat` fails, and then update Travis to make sure that we also run all tests at least once with PyOpenSSL installed. How does that sound?

benlast:
Sounds like a reasonable solution; sorry for missing that the standard tests don't have PyOpenSSL installed (is that in a doc anywhere)? I've just pushed a fix.

Lukasa:
No need to apologise!

Right now Hyper doesn't mention anything about the CI in the contributors guide, but it might be a good idea, thanks! Feel free to open an issue to track that. =) I'll wait to see how the tests go.

Lukasa:
So the tests pass, which is good, but let's try to make this report correctly. Specifically, let's try marking these tests with a [py.test directive](https://pytest.org/latest/skipping.html#marking-a-test-function-to-be-skipped) that will skip them if the SSLContext is `None`, rather than using the `if`. That way, we get a clear indication that this test isn't relevant.

benlast:
I'm not a pytest user (more of a nosetest guy), but let's see what this can do...

Lukasa:
Thanks so much for this!

PR #185, opened 2015-12-23
Minor HTTP20Connection changes

irvind:
I don't know if these tiny changes are appropriate, but here they are.

### DIFF
--- a/hyper/http20/connection.py
+++ b/hyper/http20/connection.py
@@ -722,6 +722,8 @@ class HTTP20Connection(object):
             except ConnectionResetError:
                 break
 
+            count -= 1
+
     def _send_rst_frame(self, stream_id, error_code):
         """
         Send reset stream frame with error code and remove stream from map.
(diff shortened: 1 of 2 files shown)


hyper/http20/connection.py
```diff
@@ -507,7 +507,6 @@ def _new_stream(self, stream_id=None, local_closed=False):
         """
         Returns a new stream object for this connection.
         """
-        window_size = self._settings[SettingsFrame.INITIAL_WINDOW_SIZE]
```
Lukasa: I think this being missing is a bug: looks like it should be the argument to `__wm_class()`

hyper/http20/connection.py
```diff
@@ -723,6 +722,8 @@ def _recv_cb(self):
             except ConnectionResetError:
                 break
 
+            count -= 1
+
```
Lukasa: This one looks legit though.

Lukasa:
One of these seems like it is a mistake that leads to a bug, and should be fixed properly. The other one is fixing a bug, though it'd be good to have a test for it.

irvind:
Checked.

Lukasa:
\o/ Thanks @irvind!

Issue #153, opened 2015-08-05
Hyper allows to have multiple :authority headers


direvius:
I haven't found a way to override the default ':authority' header. If I specify it in the request method, it uses both the default and specified values which causes '400 Bad Request'.
```
conn.request('GET', '/', headers={":authority": "my-hostname.org"})
```

Lukasa:
Yup, this has been something I've wanted for a long time but haven't implemented.

PR #186, opened 2015-12-23
Allow to override default request headers

irvind:
Tried to fix https://github.com/Lukasa/hyper/issues/153

### DIFF
--- a/hyper/common/headers.py
+++ b/hyper/common/headers.py
@@ -181,6 +181,28 @@ class HTTPHeaderMap(collections.MutableMapping):
         for item in self._items:
             yield item
 
+    def replace(self, key, value):
+        """
+        Replace existing header with new value. If header doesn't exist this
+        method work like ``__setitem__``. Replacing leads to deletion of all 
+        exsiting headers with the same name.
+        """
+        idx = None
+        key = to_bytestring(key)
+        for (i, (k, v)) in enumerate(self._items):
+            if _keys_equal(k, key):
+                idx = i
+                break
+
+        if idx is not None:
+            del self[key]
+
+        self._items.insert(
+            idx if idx is not None else len(self._items),
+            to_bytestring_tuple(key, value)
+        )
+        
+
     def merge(self, other):
         """
         Merge another header set or any other dict-like into this one.
--- a/hyper/common/headers.py
+++ b/hyper/common/headers.py
@@ -187,21 +187,12 @@ class HTTPHeaderMap(collections.MutableMapping):
         method work like ``__setitem__``. Replacing leads to deletion of all 
         exsiting headers with the same name.
         """
-        idx = None
-        key = to_bytestring(key)
-        for (i, (k, v)) in enumerate(self._items):
-            if _keys_equal(k, key):
-                idx = i
-                break
-
-        if idx is not None:
+        try:
             del self[key]
+        except KeyError:
+            pass
 
-        self._items.insert(
-            idx if idx is not None else len(self._items),
-            to_bytestring_tuple(key, value)
-        )
-        
+        self._items.append(to_bytestring_tuple(key, value))
 
     def merge(self, other):
         """
--- a/hyper/common/headers.py
+++ b/hyper/common/headers.py
@@ -192,7 +192,7 @@ class HTTPHeaderMap(collections.MutableMapping):
         except KeyError:
             pass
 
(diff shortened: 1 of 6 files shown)


hyper/http20/stream.py
```diff
@@ -109,11 +109,22 @@ def __init__(self,
...
+    def add_header(self, name, value, replace=False):
         """
         Adds a single HTTP header to the headers to be sent on the request.
         """
-        self.headers.append((name.lower(), value))
+        name = name.lower()
+
+        if replace:
+            try:
+                idx = [n for n, _ in self.headers].index(name)
```
Lukasa: So, this is a really unfortunate n^2 algorithm that doesn't need to be. Instead, lets change `self.headers` to be an instance of the `HTTPHeaderMap`. This will probably break some tests that do equality tests, but we can fix this by just getting `list(x.headers.items())` for anywhere that that happens.

You may want to add a `replace` method to the `HTTPHeaderMap` that first does `del` then inserts the item into `self` but that should be simple enough. This should avoid some of the duplicate code here.

Lukasa:
So for this one I think the fix works best if we move to using the HTTPHeaderMap structure defined in this project for the headers from the get-go.

irvind:
Checked.

hyper/common/headers.py
```diff
@@ -181,6 +181,28 @@ def iter_raw(self):
...
+        """
+        idx = None
+        key = to_bytestring(key)
+        for (i, (k, v)) in enumerate(self._items):
+            if _keys_equal(k, key):
+                idx = i
+                break
+
+        if idx is not None:
+            del self[key]
```
Lukasa: Given that we saved the index off, I think we can work on `_items` directly and just remove that index. That should save us from an extra loop over the container.

Lukasa: I think we can safely replace this entire pair of blocks with:

```python
try:
    del self[key]
except KeyError:
    pass
```

hyper/http20/connection.py
```diff
@@ -170,8 +170,9 @@ def request(self, method, url, body=None, headers={}):
         """
         stream_id = self.putrequest(method, url)
 
+        default_headers = (':method', ':scheme', ':authority', ':path')
         for name, value in headers.items():
-            self.putheader(name, value, stream_id)
+            self.putheader(name, value, stream_id, replace=name in default_headers)
```
Lukasa: I'm worried here about the possibility that people may pass the header as a non-native string on the platform (e.g. `unicode` on Python 2 or `bytes` on Python 3). I think we may need to do some coercion of these headers, maybe using a method like [`to_native_string` from requests](https://github.com/kennethreitz/requests/blob/master/requests/utils.py#L690-L705).

hyper/http20/stream.py
```diff
@@ -268,8 +272,11 @@ def open(self, end):
         The `end` flag controls whether this will be the end of the stream, or
         whether data will follow.
         """
+        headers = list(self.headers.items())
```
Lukasa: There should be no need for this conversion: the HTTPHeaderMap iterates like a list of two-tuples.

Lukasa:
Thanks for this, this is looking better! Some more feedback inline. =)

irvind:
@Lukasa thanks for code review. Check this out :)

hyper/common/headers.py
```diff
@@ -181,6 +181,19 @@ def iter_raw(self):
...
+        Replace existing header with new value. If header doesn't exist this
+        method work like ``__setitem__``. Replacing leads to deletion of all 
+        exsiting headers with the same name.
+        """
+        try:
+            del self[key]
+        except KeyError:
+            pass
+
+        self._items.append(to_bytestring_tuple(key, value))
```
Lukasa: At this point, I actually think we can make `replace` a simple synonym for `del` then `__setitem__`. That is, we can replace the inlined call to `__setitem__` (that is, `self._items.append`) with an actual call to `__setitem__` (`self[key] = value`), which should reduce our code duplication and reduce the risk of this breaking if the implementation changes later on.

Lukasa:
Alright, one more smaller note, but I think we're very close to what we need here! Well done!

irvind:
@Lukasa :ok_hand:

Lukasa:
\o/ Looks good to me! Thanks so much!

PR #188, opened 2016-01-12
Handle errors better

fredthomsen:
Nowhere near ready, but wanted to know if I was heading down the right path...


Lukasa:
@fredthomsen So this is a great start, but I think it might not be the best way to go about it. I suspect the better way to do this is for me to migrate the core of hyper to use hyper-h2, which has much more effective error handling in it.

fredthomsen:
Woops, I misunderstood earlier then. My fault. I will leave this alone for now then.

PR #193, opened 2016-01-25
Add support of AppEngine SSL library

jmuk:
This is a patch to provide the SSL comptibility module working within AppEngine. This change does not enable hyper working on AppEngine yet, but I think this would help approaching to it.

### DIFF
--- a/hyper/compat.py
+++ b/hyper/compat.py
@@ -36,11 +36,11 @@ def ignore_missing():
         pass
 
 if is_py2:
-    if is_py2_7_9_or_later:
-        import ssl
-    elif is_appengine:
+    if is_appengine:
         from . import ssl_compat_appengine
         ssl = ssl_compat_appengine
+    elif is_py2_7_9_or_later:
+        import ssl
     else:
         ssl = ssl_compat
 
--- a/hyper/compat.py
+++ b/hyper/compat.py
@@ -15,11 +15,18 @@ except ImportError:
     # TODO log?
     ssl_compat = None
 
+try:
+    import google.appengine
+    is_appengine = True
+except ImportError:
+    is_appengine = False
+
 _ver = sys.version_info
 is_py2 = _ver[0] == 2
 is_py2_7_9_or_later = _ver[0] >= 2 and _ver[1] >= 7 and _ver[2] >= 9
 is_py3 = _ver[0] == 3
 is_py3_3 = is_py3 and _ver[1] == 3
+
 
 @contextmanager
 def ignore_missing():
@@ -31,6 +38,9 @@ def ignore_missing():
 if is_py2:
     if is_py2_7_9_or_later:
         import ssl
+    elif is_appengine:
+        from . import ssl_compat_appengine
+        ssl = ssl_compat_appengine
     else:
         ssl = ssl_compat
 
--- a/hyper/http20/connection.py
+++ b/hyper/http20/connection.py
@@ -10,6 +10,7 @@ from ..common.exceptions import ConnectionResetError
 from ..common.bufsocket import BufferedSocket
 from ..common.headers import HTTPHeaderMap
 from ..common.util import to_host_port_tuple, to_native_string
+from ..compat import is_appengine
 from ..packages.hyperframe.frame import (
     FRAMES, DataFrame, HeadersFrame, PushPromiseFrame, RstStreamFrame,
     SettingsFrame, Frame, WindowUpdateFrame, GoAwayFrame, PingFrame,
@@ -254,7 +255,10 @@ class HTTP20Connection(object):
                 proto = H2C_PROTOCOL
 
             log.debug("Selected NPN protocol: %s", proto)
-            assert proto in H2_NPN_PROTOCOLS or proto == H2C_PROTOCOL
+            # AppEngine SSLSocket does not have selected_npn_protocol nor
+            # selected_alpn_protocol, so just avoid this assertion.
+            assert is_appengine or (
+                proto in H2_NPN_PROTOCOLS or proto == H2C_PROTOCOL)
 
             self._sock = BufferedSocket(sock, self.network_buffer_size)
 
--- a/hyper/http20/connection.py
+++ b/hyper/http20/connection.py
@@ -10,7 +10,6 @@ from ..common.exceptions import ConnectionResetError
 from ..common.bufsocket import BufferedSocket
 from ..common.headers import HTTPHeaderMap
 from ..common.util import to_host_port_tuple, to_native_string
-from ..compat import is_appengine
 from ..packages.hyperframe.frame import (
     FRAMES, DataFrame, HeadersFrame, PushPromiseFrame, RstStreamFrame,
     SettingsFrame, Frame, WindowUpdateFrame, GoAwayFrame, PingFrame,
@@ -255,10 +254,7 @@ class HTTP20Connection(object):
                 proto = H2C_PROTOCOL
 
             log.debug("Selected NPN protocol: %s", proto)
-            # AppEngine SSLSocket does not have selected_npn_protocol nor
-            # selected_alpn_protocol, so just avoid this assertion.
-            assert is_appengine or (
-                proto in H2_NPN_PROTOCOLS or proto == H2C_PROTOCOL)
+            assert proto in H2_NPN_PROTOCOLS or proto == H2C_PROTOCOL
 
             self._sock = BufferedSocket(sock, self.network_buffer_size)
 
--- a/hyper/ssl_compat_appengine.py
+++ b/hyper/ssl_compat_appengine.py
@@ -19,7 +19,7 @@ from ssl import match_hostname
 OP_NO_COMPRESSION = 0
 
 
-class SSLContext:
+class SSLContext(object):
     """A SSL context which implements the methods used by hyper."""
 
     def __init__(self, proto):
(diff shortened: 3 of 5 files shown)


hyper/ssl_compat_appengine.py
```diff
@@ -0,0 +1,49 @@
...
+import ssl
+
+from ssl import PROTOCOL_TLSv1, PROTOCOL_SSLv23, PROTOCOL_SSLv3
+from ssl import CERT_NONE, CERT_OPTIONAL, CERT_REQUIRED
+from ssl import match_hostname
+
+OP_NO_COMPRESSION = 0
+
+
+class SSLContext:
```
Lukasa: This is a minor thing, but can we explicitly inherit from `object` here? It's unnecessary with Python 3, but old-style classes are a nasty surprise when working in Python 2 code and so it'd be good if we could save ourselves the hassle.

hyper/ssl_compat_appengine.py
```diff
@@ -0,0 +1,49 @@
...
+This module complements some constants and classes which don't exist on
+AppEngine's SSL module, to be used by hyper smoothly.
+See: https://cloud.google.com/appengine/docs/python/sockets/ssl_support
+"""
+
+import ssl
+
+from ssl import PROTOCOL_TLSv1, PROTOCOL_SSLv23, PROTOCOL_SSLv3
+from ssl import CERT_NONE, CERT_OPTIONAL, CERT_REQUIRED
+from ssl import match_hostname
```
Lukasa: Does AppEngine promise to provide `match_hostname`?

hyper/ssl_compat_appengine.py
```diff
@@ -0,0 +1,49 @@
...
+    def wrap_socket(self, sock, server_side=False, do_handshake_on_connect=True,
+                    suppress_ragged_eofs=True, server_hostname=None):
+        sock = ssl.wrap_socket(
+            sock, server_side=server_side, cert_reqs=self.verify_mode,
+            ssl_version=self.protocol_version, ca_certs=self.custom_ca_cert,
+            suppress_ragged_eofs=suppress_ragged_eofs,
+            do_handshake_on_connect=do_handshake_on_connect)
+        # AppEngine SSLSocket does not have selected_npn_protocol, a hacky
+        # solution to return a dummy data.
+        sock.selected_npn_protocol = lambda: 'h2c'
```
Lukasa: Is `h2c` a deliberate choice here? If so, what are you aiming to suggest?

Lukasa:
Thanks for this @jmuk!

I've left a bit of inline feedback, but I'm interested in getting an idea of how much time you have to spend here. It'd be really great if we could add some tests to validate that all of this logic works correctly, but as you've pointed out, `hyper` doesn't directly work with GAE yet even with this patch in place. Are you expecting multiple patches to come down the line shortly to add that function? If you are, I'd be interested in having these PRs land in a parallel branch to the mainline development so that we can get them all in place and then add tests in one fell swoop before merging to mainline.

jmuk:
Thank you for the quick reply.

I'll provide the fixes on your comments, but for the questions; I'm good to be within a separate branch. I actually have another CL for the usage of socket.recv_into (which is not implemented yet), but that is all I've noticed. after that hyper will work on GAE -- I mean, at least it can behave as a http2 client in the way I tried, some may be missing though.

hyper/ssl_compat_appengine.py
jmuk: well, I'm afraid not, but it's available actually as far as I tried, and I think there's not a big reason to drop it now.
I can put try: with catching ImportError, but in that case I'm not sure how the fake match_hostname should be implemented.

hyper/ssl_compat_appengine.py
jmuk: I didn't choose it deliberately -- I just wanted to avoid the assertion in connection.
After seeing your comment, I changed my mind and put a condition to the assertion which is used.
Note that I don't care much about http11 module, mostly on GAE developers will use runtime-provided url_fetch anyways.

hyper/ssl_compat_appengine.py
jmuk: Done

hyper/ssl_compat_appengine.py
Lukasa: This was mostly a question about whether GAE provides this function in general. If it does, I'm not too worried about having it as an unconditional import.

hyper/http20/connection.py
```diff
@@ -254,7 +255,10 @@ def connect(self):
                 proto = H2C_PROTOCOL
 
             log.debug("Selected NPN protocol: %s", proto)
-            assert proto in H2_NPN_PROTOCOLS or proto == H2C_PROTOCOL
+            # AppEngine SSLSocket does not have selected_npn_protocol nor
+            # selected_alpn_protocol, so just avoid this assertion.
+            assert is_appengine or (
+                proto in H2_NPN_PROTOCOLS or proto == H2C_PROTOCOL)
```
Lukasa: Sorry, I think I wasn't clear enough with my previous feedback. When I was asking about `H2C_PROTOCOL` I wanted to know why you chose that over `h2`. `h2c` is meant to be reserved for HTTP/2 via opportunistic encryption, which seemed an odd choice. I'd be happy to use `h2`.

jmuk: Okay, pushed a new patch -- now it uses `'h2'`

jmuk:
Added test cases, this will make travis green.

test/test_appengine.py
```diff
@@ -0,0 +1,60 @@
...
+        context.verify_mode = ssl_compat_appengine.CERT_NONE
+        context.verify_hostname = False
+        # This invocation makes sure it does not fail.
+        context.load_verify_locations('test/certs/server.crt')
+        sock = socket.create_connection(
+            (self.server_thread.host, self.server_thread.port))
+        ssl_sock = context.wrap_socket(sock)
+        assert ssl_sock.selected_npn_protocol() == 'h2'
+        ssl_sock.close()
+        self.tear_down()
```
Lukasa: Presumably these tests should only run if we're actually running on GAE or in a GAE-like system? Otherwise they don't really prove all that much.

I think we need to mark these test classes with `pytest.mark.skipif()` and use similar logic to what urllib3 uses for its GAE tests, and then add that to the matrix of tests we run on Travis. That might be a little tricky: do you want me to handle doing that?

Lukasa: There's also a timing issue here: the socket_handler really needs to wait until the connection is made and the assertions have been made before any of the tests run. Otherwise, you get problems like the one you can see in the travis run.

The easiest way to do this is to use threading events to synchronise the activity.

Lukasa:
Your Python 3.2 test failures are also hitting an issue that I just resolved in #194. You may want to rebase on top of `development` in order to obtain the fix.

test/test_appengine.py
jmuk: GAE doesn't offer a different ssl library actually, it's the same ssl as bundled in future version of Python, but the values in the ssl module is slightly limited. SSLContext doesn't exist, for example.

That means that this ssl_compat_appengine should work with the normal ssl module outside of AppEngine (if it's newer than 2.7.9), and the test case would be still useful to verify the behavior of the module itself.

For the second comment, thank you for catching that. I put sock.do_handshake() on socket_handler so that it verifies to establish the connection, so I think that's good enough.

Lukasa: Ah, ok, if this should work outside GAE then it's probably enough to just add an extra test environment to validate that it *also* works on GAE. If we're going to support GAE then we should support it properly! =D

jmuk: Well, I'm afraid it's not possible to verify if it works on GAE here on travisCI. Actually the local emulator (dev_appserver) in AppEngine SDK simply uses the local runtime with local ssl library, so that doesn't mean the actual GAE environment.

Google could put a test case in its internal end-to-end test cases if hyper is listed in the bundled third party libraries (https://cloud.google.com/appengine/docs/python/tools/libraries27?hl=en), but that's anyways out of the scope right now.

jmuk:
It seems I've forgotten to reply, but yeah, I've rebased to the top of `development`, and now it succeeds.

Is there anything I should do?

test/test_appengine.py
Lukasa: Aha, ok, fair enough. In that case, let me do a final code review now. =)

hyper/compat.py
```diff
@@ -31,6 +38,9 @@ def ignore_missing():
 if is_py2:
     if is_py2_7_9_or_later:
         import ssl
+    elif is_appengine:
+        from . import ssl_compat_appengine
+        ssl = ssl_compat_appengine
```
Lukasa: Is this order of fallbacks correct? It seems like if GAE identifies itself as Python 2.7.9 or later then we won't use the AppEngine compatibility module. Having not used GAE myself I'm not sure if this order is right...

Lukasa:
M'kay, I have one smaller question about the order of fallbacks.

hyper/compat.py
jmuk: Fixed, use ssl_compat_appengine when appengine.

When an AppEngine runtime got updated to Python 2.7.9 or later, I was personally thinking that it will use the bundled ssl and this compatibility module becomes unnecessary.
Indeed, however, nothing is sure for now, and anyways this would work with bundled ssl as I commented on another place, using it for GAE environment would be at least harmless even in that.

jmuk:
Sorry, I've come to realize that the lack of npn/alpn protocol could be significant, and adding them can't be done in hyper -- ssl library (in appengine runtime) needs to be updated. That means, unfortunately this current patch is insufficient, and some details in this PR should be changed when ssl library is updated.

Let me close this PR for the time being. Thank you for your reviews.

PR #194, opened 2016-01-26
Make passphrase a bytestring.

Lukasa:
This should resolve some Python 3.2 test breakages.


Lukasa:
In the longer term we may simply want to drop support for Python 3.2.

PR #197, opened 2016-02-11
Allow protocol version to be specified, useful if ALPN is unavailable

markjenkins:
I needed this when connecting to the new HTTP/2 Apple Push Notification Service with python 2.7.11 on a platform where the openssl support available did not include ALPN (ssl.HAS_ALPN was False, ssl.HAS_NPN is True).

The Apple Push Notification Service doesn't allow negotiation by NPN, only ALPN works, so the following assertion in hyper/http20/connection.py:
`assert proto in H2_NPN_PROTOCOLS or proto == H2C_PROTOCOL`
would fail for me as proto would end up None as a result of wrap_socket in hyper/tls.py expecting either NPN or ALPN to work.

I was able to determine elsewhere with a working ALPN implementation that Apple Push Notification Service protocol is always 'h2'. So the point of this change is to allow a user who knows what the protocol will be to override, I supply 'h2' for the force_proto argument provided here and things work.

### DIFF
--- a/hyper/http20/connection.py
+++ b/hyper/http20/connection.py
@@ -66,7 +66,8 @@ class HTTP20Connection(object):
         and one also isn't provided in the ``proxy`` parameter, defaults to 8080.
     """
     def __init__(self, host, port=None, secure=None, window_manager=None, enable_push=False,
-                 ssl_context=None, proxy_host=None, proxy_port=None, **kwargs):
+                 ssl_context=None, proxy_host=None, proxy_port=None,
+                 force_proto=None, **kwargs):
         """
         Creates an HTTP/2 connection to a specific server.
         """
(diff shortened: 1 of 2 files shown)


Lukasa:
Mark, just to clarify: APNS will negotiate via ALPN but, if ALPN is not supported, assumes HTTP/2? That's...unexpected.

Lukasa:
Well that *is* unexpected. I think the general idea here is good, but when I have a bit more time later today I may provide more detailed review.

Lukasa:
Thanks for this @markjenkins, and I'm sorry I let it languish for so long! Because I did a really big refactor of the codebase while this was languishing I went ahead and merged it manually, but I'm really grateful for the change. Thanks so much!

Issue #152, opened 2015-08-05
Impossible to override default headers from CLI


direvius:
Default headers are started with ':' and it is used to separate headers key and value in CLI arguments. And hyper does not add ':' to the specified key also. For example, hyper sets ```:authority``` to the hostname by default and you can't override it:
```
$ hyper --debug GET https://google.com:443/ authority:google.ru
...
HPACK encoding [(':method', 'GET'), (':scheme', 'https'), (':authority', 'google.com'), (':path', '/'), ('authority', 'google.ru')]
```
Or even this way:
```
$ hyper --debug GET https://google.com:443/ :authority:google.ru
...
HPACK encoding [(':method', 'GET'), (':scheme', 'https'), (':authority', 'google.com'), (':path', '/'), ('', 'authority:google.ru')]
```

Lukasa:
Agreed, this needs working on too. =)

(The reason I'm being somewhat non-committal on these reports is because I'm currently on holiday, so I don't have a *ton* of time to work on hyper at the moment. That said, there are definitely things that need to get done.)

PR #200, opened 2016-02-18
issue #152 bugfix

pkrolikowski:
Override default headers is now possible.

### DIFF
--- a/hyper/cli.py
+++ b/hyper/cli.py
@@ -164,7 +164,13 @@ def set_request_data(args):
     body, headers, params = {}, {}, {}
     for i in args.items:
         if i.sep == SEP_HEADERS:
-            headers[i.key] = i.value
+           # key:value case
+           if i.key != '':
+               headers[':' + i.key] = i.value
+           # :key:value case
+           else:
+               k, v = i.value.split(':')
+               headers[':' + k] = v
         elif i.sep == SEP_QUERY:
             params[i.key] = i.value
         elif i.sep == SEP_DATA:
--- a/hyper/http20/connection.py
+++ b/hyper/http20/connection.py
@@ -175,6 +175,8 @@ class HTTP20Connection(object):
 
         default_headers = (':method', ':scheme', ':authority', ':path')
         for name, value in headers.items():
+            # rewrite to default headers format
+            name = ':' + name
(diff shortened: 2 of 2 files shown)


Lukasa:
Hi @pkrolikowski! Thanks for proposing this change! I'm looking forward to working through it with you!

So I don't believe this is addressing the core issue correctly. The problem here specifically is that the way the CLI handles the headers does not allow for overriding the special headers (like `:authority`). This is not a problem in the `HTTP20Connection` class, which *does* allow it (and is where you've made a change). Instead, the problem is likely to be in the `cli.py` file somewhere.

Take a look at this suggested call to the CLI tool:

```
$ hyper --debug GET https://google.com:443/ :authority:google.ru
...
HPACK encoding [(':method', 'GET'), (':scheme', 'https'), (':authority', 'google.com'), (':path', '/'), ('', 'authority:google.ru')]
```

This suggests that the CLI tool is setting the header to `('', 'authority:google.ru')`. This is probably a parsing problem: we'd like to have a leading colon if possible.

pkrolikowski:
OK, I'll look at cli.py. So the correct behave should be like this:

$ hyper --debug GET https://google.com:443/ :authority:google.ru
...
HPACK encoding [(':method', 'GET'), (':scheme', 'https'), (':authority', 'google.com'), (':path', '/'), (':', 'authority:google.ru')]

$ hyper --debug GET https://google.com:443/ authority:google.ru
...
HPACK encoding [(':method', 'GET'), (':scheme', 'https'), (':authority', 'google.ru'), (':path', '/')]

Am I right ?

Lukasa:
No, I think the correct behaviour should be:

```
$ hyper --debug GET https://google.com:443/ :authority:google.ru
...
HPACK encoding [(':method', 'GET'), (':scheme', 'https'), (':authority', 'google.ru'), (':path', '/')]
```

pkrolikowski:
I've changed the way how set_request_data() in cli.py is processing headers separation.

Lukasa:
Thanks for this @pkrolikowski! That code looks plausible enough to me.

To verify that it works, can you please write some tests to see how it functions with a variety of inputs? (I apologise for the Travis problems, it seems that Travis has a timing issue with our disabling TLS certificate verification in the tests)

PR #205, opened 2016-02-22
Enable override default headers from CLI

pkrolikowski:
@Lukasa I had a problem with push to rebased branch, so I decided to make a new request.

### DIFF
--- a/hyper/cli.py
+++ b/hyper/cli.py
@@ -164,12 +164,14 @@ def set_request_data(args):
     body, headers, params = {}, {}, {}
     for i in args.items:
         if i.sep == SEP_HEADERS:
-            # :key:value case
-            if i.key == '':
-                k, v = i.value.split(':')
-                headers[':' + k] = v
-            else:
+            if i.key:
                 headers[i.key] = i.value
+            else:
+            # when overriding a HTTP/2 special header there will be a leading 
+            # colon, which tricks the command line parser into thinking 
+            # the header is empty
+                k, v = i.value.split(':', 1)
+                headers[':' + k] = v
         elif i.sep == SEP_QUERY:
             params[i.key] = i.value
         elif i.sep == SEP_DATA:
--- a/hyper/cli.py
+++ b/hyper/cli.py
@@ -164,7 +164,12 @@ def set_request_data(args):
     body, headers, params = {}, {}, {}
     for i in args.items:
         if i.sep == SEP_HEADERS:
-            headers[i.key] = i.value
+            # :key:value case
+            if i.key == '':
+                k, v = i.value.split(':')
+                headers[':' + k] = v
+            else:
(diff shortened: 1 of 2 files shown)


hyper/cli.py
```diff
@@ -164,7 +164,12 @@ def set_request_data(args):
     body, headers, params = {}, {}, {}
     for i in args.items:
         if i.sep == SEP_HEADERS:
-            headers[i.key] = i.value
+            # :key:value case
+            if i.key == '':
+                k, v = i.value.split(':')
```
Lukasa: While I don't think it's likely we could get a header with a colon in it (the only possible one is `:path`, which *might* be able to have a colon in it), we should defend against that anyway. Let's swap this to `i.value.split(':', 1)`.

hyper/cli.py
```diff
@@ -164,7 +164,12 @@ def set_request_data(args):
     body, headers, params = {}, {}, {}
     for i in args.items:
         if i.sep == SEP_HEADERS:
-            headers[i.key] = i.value
+            # :key:value case
+            if i.key == '':
```
Lukasa: Generally speaking, good Python style tests for the empty string by saying `if not i.key`, rather than explicitly comparing to the empty string. For that reason, let's flip this entire block on its head, changing it to:

```python
if i.key:
    headers[i.key] = i.value
else:
    # :key:value case
    k, v = i.value.split(':', 1)
    headers[':' + k] = v
```

Let's also make this comment a little bit more expressive: something like "if the user wants to override a HTTP/2 special header there will be a leading colon, which tricks the command line parser into thinking the header is empty".

Lukasa:
Fantastic, everything is green! I have a few minor style notes that I'd like you to change, if possible, but when that's done I think this will be ready to merge.

pkrolikowski:
I changed this block as you suggested.

hyper/cli.py
```diff
@@ -164,7 +164,14 @@ def set_request_data(args):
     body, headers, params = {}, {}, {}
     for i in args.items:
         if i.sep == SEP_HEADERS:
-            headers[i.key] = i.value
+            if i.key:
+                headers[i.key] = i.value
+            else:
+            # when overriding a HTTP/2 special header there will be a leading 
+            # colon, which tricks the command line parser into thinking 
+            # the header is empty
```
Lukasa: Mind indenting this one more level?

Lukasa:
Sorry, I wasn't quite clear enough on the indentation. Fix that up and we're all good! =)

Lukasa:
Magnificent, thank you so much @pkrolikowski! :sparkles: :cake: :sparkles:

pkrolikowski:
I am glad I could help :) So what do you think, what could I do next ?

PR #208, opened 2016-03-02
Add content-length header for http2 connections

### LINKED ISSUE #206: Problem sending JSON data as request body
I am trying to do some testing using the http2 version of httbin: https://nghttp2.org/httpbin/. Posting to the url `https://nghttp2.org/httpbin/post` will respond with a JSON object containing information about the request. This JSON includes the body and data sent with the request. Using hyper, it appears that the data that I'm sending as the request body is not being sent. Here's an example:

	from hyper import HTTPConnection
	conn = HTTPConnection('nghttp2.org:443')

	req = conn.request(
		'POST',
		'/httpbin/post',
		body=b'{"foo": "bar"}',
	)

	resp = conn.get_response()
	print(resp.read())

Using a different http2 client, I get the expected response:

	{
	  "args": {},
	  "data": "{\"foo\": \"bar\"}",  <-- critical bit right here
	  "files": {},
	  "form": {},
	  "headers": {
		"Accept": "*/*",
		"Accept-Encoding": "gzip, deflate",
		"Accept-Language": "en-US,en;q=0.8",
		"Cache-Control": "no-cache",
		"Content-Length": "14",
		"Content-Type": "application/json",
		"Csp": "active",
		"Host": "nghttp2.org",
		"Origin": "chrome-extension://aicmkgpgakddgnaphhhpliifpcfhicfo",
		"Postman-Token": "70fa24f0-9e35-82ef-c838-7c15461ba3bd",
		"User-Agent": "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Ubuntu Chromium/43.0.2357.130 Chrome/43.0.2357.130 Safari/537.36",
		"Via": "2 nghttpx"
	  },
	  "json": {
		"foo": "bar"
	  },
	  "origin": "68.180.88.224",
	  "url": "https://nghttp2.org/httpbin/post"
	}

But using the python given above, I get the response: 

	{
	  "args": {},
	  "data": "",
	  "files": {},
	  "form": {},
	  "headers": {
		"Host": "nghttp2.org",
		"Transfer-Encoding": "chunked",
		"Via": "2 nghttpx"
	  },
	  "json": null,
	  "origin": "68.180.88.224",
	  "url": "https://nghttp2.org/httpbin/post"
	}

I'm using version 0.5.0 of hyper, Python 3.4.4. Am I doing something wrong with my request, or is there a way to get more verbose output from hyper so that I can see what's really going on?

natemara:
If a body is defined, we should be sending a Content-Length header unless the request has the Transfer-Encoding header is set. This should fix #206. I was having some difficulty running the tests, but it seems like that was mostly related to SSL certificate problems. Looking at [the spec doc](http://httpwg.org/specs/rfc7230.html#header.content-length), it says that clients SHOULD send the Content-Length header when there is no Transfer-Encoding header set.


Lukasa:
Thanks for this @natemara! The test troubles are interesting, I've got a suspicion they're timing related, but I'll try to take a look and see if I can fix them so that we can write some tests for this.

Lukasa:
Ok, so I can't reproduce your testing problems locally, and I believe I fixed all the relevant problems in #202. Does your local branch contain #202?

natemara:
Yes, I'm working on the development branch, with up to #205.

Lukasa:
Hmm, that's perplexing. What's your environment? What Python are you using, what OpenSSL etc.?

Lukasa:
@natemara Do you want to try rebasing onto the current development branch and try again?

natemara:
Alright, that's rebased and checking now.

Lukasa:
Ok, so this is doing better. There are two things that need to be solved:

- we need to update the tests (see the Python 2 result for some problems)
- the function doesn't work as expected on Python 3 (for boring string type reasons)

Both of those will need to be fixed up. =) You up for doing that @natemara?

hyper/http20/connection.py
```diff
@@ -170,6 +170,8 @@ def request(self, method, url, body=None, headers={}):
         # Convert the body to bytes if needed.
         if body and isinstance(body, (unicode, bytes)):
             body = to_bytestring(body)
+            if 'Transfer-Encoding' not in headers.keys():
```
Hellowlol: Any reason why we need O(n) here? Would it be better to just use:
```
if 'Transfer-Encoding' not in headers:
    ....

```

PR #212, opened 2016-03-18
Depend on hyper-h2.

Lukasa:
This PR tracks the work to move to depend on hyper-h2 instead of using the HTTP/2 implementation baked into this tool.


Lukasa:
Ok, I'm just going to go ahead and merge this. I don't think this is a suitable replacement for a full refactor, but it's a good first step. Sitting on it for longer won't make anything any easier.

PR #217, opened 2016-04-08
HTTPConnection takes a None default value for headers, instead of {}

AvivC:
Pycharm adds an `.idea/` folder which shouldn't be pushed to the repo.

### DIFF
--- a/hyper/common/connection.py
+++ b/hyper/common/connection.py
@@ -76,7 +76,7 @@ class HTTPConnection(object):
             self._host, self._port, **self._h1_kwargs
         )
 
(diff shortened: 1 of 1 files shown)


Lukasa:
Generally speaking, I'm disinclined to want to include IDE-specific files in a project-specific gitignore. If your IDE generates files that should be excluded from the repo, you should use a [global gitignore file](https://help.github.com/articles/ignoring-files/).

AvivC:
I see. I changed the PR to include a single commit, just the last one.

Lukasa:
Thanks for this! :sparkles: :cake: :sparkles:

PR #218, opened 2016-04-08
Default value for headers arg in HTTP20Connection.request() is None instead of {}

AvivC:


Lukasa:
Thanks! :sparkles: :cake: :sparkles:

AvivC:
Sure

PR #220, opened 2016-04-11
Use the right alphabet for HTTP2-Settings (fixes #219)

### LINKED ISSUE #219: Malformed HTTP2-Settings
Hyper produces the HTTP2-Settings header as follows:

    headers[b'HTTP2-Settings'] = base64.b64encode(http2_settings.serialize_body())

This uses the default Base64 alphabet with padding. But [RFC 7540 § 3.2.1](https://tools.ietf.org/html/rfc7540#section-3.2.1) requires the [URL- and filename-safe Base64 alphabet](https://tools.ietf.org/html/rfc4648#section-5) with padding removed. So it should look something like this:

    settings_bytes = http2_settings.serialize_body()
    headers[b'HTTP2-Settings'] = base64.b64encode(settings_bytes,
                                                  b'-_').rstrip(b'=')

Server implementations seem to Postel this out, but I think correctness won’t hurt here.

vfaronov:


Lukasa:
Thanks for this! :sparkles: :cake: :sparkles:

Issue #187, opened 2015-12-24
Timeout handling


farefernandez:
Hi,

I'm using Hyper to connect to a HTTP/2 website and I'd like to be able to control timeouts (connection, read, dns resolver).

I don't seem to be able to find mentions to timeout handling in the documentation so I assume this could be seen as a feature request.

Thanks!

Lukasa:
Yup, this seems like a good feature request to me!

farefernandez:
Thanks Cory.

API-wise, we could have one global, coarse-grained timeout parameter which could refer to the sum of all of the components (connect, read, dns). This is what people usually should use.

Fine-grained timeouts could be added with specific parameters controlling each of the components.

farefernandez:
For reference, this is how requests does it (although the dns resolver timeout doesn't seem to be handled): http://docs.python-requests.org/en/latest/user/advanced/#timeouts

Relevant DNS resolver discussion: https://bugs.python.org/issue22889

Thanks!

Lukasa:
@farefernandez I'm a bit nervous about implementing an extremely complex timeout logic. However, if you were going in that direction the thing to do is to look at urllib3's `Timeout` logic and see if we can implement something somewhat similar.

johnwheeler:
Hi All,

I'm getting this error consistently on response.read() 

```
ssl.SSLError: ('The read operation timed out',)
```

Does this issue mean that this can't be handled currently?

Thank you for hyper!

Lukasa:
@johnwheeler I think that depends on what you mean by "handled". What would you define as handled here?

johnwheeler:
Thank you for the quick response. Handled was a poor word choice since it connotes exception handling. I was wondering if there's a way to stop this error from occurring and proceed to read the data, and I want to confirm that:

1. There's no way to do that per this issue (no way to set a timeout)
2. I'm not misunderstanding the HTTP/2 protocol and this error makes it immediately apparent to you how.

Thanks again

johnwheeler:
Please disregard. I just don't know what I'm doing. I think I'm trying to interact with a HTTP/2 service in an HTTP/1 way and didn't make it far enough in your docs. I've gotten further along now, and it looks like I'm making headway. 

FYI - I'm using this library to connect to the Alexa Voice Service, which is Amazon's Echo service for building the Echo into your own hardware.

Lukasa:
Ah I'm glad you've worked it through!

johnwheeler:
[...]

PR #221, opened 2016-04-12
Improve handling of socket errors

tbetbetbe:
- Use the default connection timeout set by socket.setdefaulttimeout().

* I.e, don't specify one, until there is [an API][1] for timeouts, it's
probably better for libraries to be able to set the socket timeout
outside of hyper via [setdefaulttimeout][2]

- Updates error handling to simulate blocking behaviour when the socket
  is in timeout mode (which is internally a kind of non-blocking
  mode). Specifically, during reads, add
  * handling of specific ssl.SSLErrors WANT_READ and WANT_WRITE
  * handling of errno.EAGAIN and errno.EINTR

[1]: https://github.com/Lukasa/hyper/issues/187
[2]: https://docs.python.org/2/library/socket.html#socket.setdefaulttimeout

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2014 Cory Benfield
+Copyright (c) 2014 Cory Benfield, Google Inc
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/hyper/http20/connection.py
+++ b/hyper/http20/connection.py
@@ -9,6 +9,7 @@ import h2.connection
 import h2.events
 import h2.settings
 
+from ..compat import ssl
 from ..tls import wrap_socket, H2_NPN_PROTOCOLS, H2C_PROTOCOL
 from ..common.exceptions import ConnectionResetError
 from ..common.bufsocket import BufferedSocket
@@ -237,7 +238,7 @@ class HTTP20Connection(object):
                 host = self.proxy_host
                 port = self.proxy_port
 
-            sock = socket.create_connection((host, port), 5)
+            sock = socket.create_connection((host, port))
 
             if self.secure:
                 assert not self.proxy_host, "Using a proxy with HTTPS not yet supported."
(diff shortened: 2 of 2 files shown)


Lukasa:
Firstly, thanks @tbetbetbe! This is a good set of changes.

I have some concerns about the branch for ssl errors though. Specifically, I'm concerned about repeatedly looping on `ssl.SSL_ERROR_WANT_WRITE` when we're only reading from the socket. I'm concerned that, if that's raised, we could spin forever there.

Does the python ssl module do something to avoid that being a risk?

tbetbetbe:
> Does the python ssl module do something to avoid that being a risk?

Not that I'm aware of.

The PR as is addresses a set of sporadic but consistent test failures I was seeing in a library I'm working on that's based on hyper.

ssl.SSL_ERROR_WANT_WRITE should only get raised if something is writing to the socket when it's in timeout mode.  I believe it's [common](http://www.programcreek.com/python/example/13216/socket.SSL_ERROR_WANT_WRITE) to treat it as a transient failure here, on the assumption that at this point it's not a handshake error and any concurrent writer seeing that transient failure will handle it by continuing.  I.e, the approach is effective if the potential readers and writer of the socket are all handling the transient errors in the same way.

Because of that, it's may be a good idea to replace the call to sendall with a loop around send that does the same kind of handling.  At the moment, if sendall fails with a transient SSL error, recovery will not be possible as we don't know how much data was written.

Another approach is to place a timeout on the loop as in [ssl_compat](https://github.com/Lukasa/hyper/blob/998aa02e7b94735ea7a7799f2e84fa064aaaa4ac/hyper/ssl_compat.py#L99) 

I'd be happy to either 
- update this PR with timeout for the loop or
- update this PR so that it changes sendall as well or
- leave this change as is and track the sendall/timeout changes as a proposal in a separate issue or
- switch this change to loop on ssl.SSL_ERROR_WANT_READ as you suggest 

Any of these are OK, as long the tests failures I was seeing remain resolved.  Which approach to you prefer ?

Lukasa:
I think just merging this is fine: consider me swayed by your argument. =D

matangover:
If I'm not mistaken, this PR has introduced a NameError in the code - please observe [line 545](https://github.com/Lukasa/hyper/pull/221/files#diff-569343287fd890f6fd59068d809b1d92R545) in `connection.py` (it accesses `sleep`, which is not imported).

Thanks goes to pylint for warning about this. :)

tbetbetbe:
Yup, sorry that's correct.

That should be `time.sleep`, with a corresponding import of the time module.  I will prepare a follow-up PR to correct this.

Lukasa:
Thanks both @matangover and @tbetbetbe!

Some day this week I'll be cleaning up the linter errors in this project and enforcing flake8 compliance in the CI suite: consider yourselves warned. ;)

Lukasa:
@tbetbetbe There's no need for a follow-up PR: I've started work on the linting today, and so I'll be sweeping it up anyway. =)

Lukasa:
See #226.

PR #223, opened 2016-04-16
Validate http11 body type

AvivC:
Raising `ValueError` in `HTTP11Connection._send_body` on invalid body type

Added tests for the case when the error may be raised. Also testing the
other cases when errors regarding the type of `body` argument may be
raised.
Added to the docstring the option to send an iterable of bytestrings as a body.

### DIFF
--- a/test/test_http11.py
+++ b/test/test_http11.py
@@ -269,13 +269,14 @@ class TestHTTP11Connection(object):
     def test_content_length_overrides_generator(self):
         c = HTTP11Connection('httpbin.org')
         c._sock = sock = DummySocket()
+
         def body():
             yield b'hi'
             yield b'there'
             yield b'sir'
 
         c.request(
-            'POST', '/post', headers={b'content-length': b'10'}, body=body()
+            'POST', '/post', body=body(), headers={b'content-length': b'10'}
         )
 
         expected = (
@@ -288,8 +289,8 @@ class TestHTTP11Connection(object):
             b"\r\n"
             b"hitheresir"
         )
-        received = b''.join(sock.queue)
 
+        received = b''.join(sock.queue)
         assert received == expected
 
     def test_chunked_overrides_body(self):
@@ -414,6 +415,7 @@ class TestHTTP11Connection(object):
     def test_content_length_overrides_generator_unicode(self):
         c = HTTP11Connection('httpbin.org')
         c._sock = DummySocket()
+
         def body():
             yield u'hi'
             yield u'there'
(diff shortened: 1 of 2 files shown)


test/test_http11.py
```diff
@@ -446,6 +448,38 @@ def test_http_upgrade_headers_only_sent_once(self):
...
+            c.request('GET', '/get', body=body, headers={'content-length': str(len(str(body)))})
+        assert str('Request body must be a bytestring, a file-like object returning bytestrings '
+                   'or an iterable of bytestrings. Got: {}'.format(type(body))) in str(exc_info)
+
+        rouge_element = 123
+        with pytest.raises(TypeError) as exc_info:
+            body = ['legal1', 'legal2', rouge_element]
+            c.request('GET', '/get', body=body, headers={'content-length': str(len(map(str, body)))})
+        assert str('Elements in iterable body must be bytestrings. Illegal element: {}'
+                   .format(rouge_element)) in str(exc_info)
```
Lukasa: Can these be split into two separate tests, please?

test/test_http11.py
```diff
@@ -446,6 +448,38 @@ def test_http_upgrade_headers_only_sent_once(self):
...
+
+    def test_exception_raised_for_filelike_body_not_returning_bytes(self):
+        c = HTTP11Connection('httpbin.org')
+
+        class RougeFile(object):
```
Lukasa: Typo: you want `Rogue` here. =)

test/test_http11.py
```diff
@@ -446,6 +448,38 @@ def test_http_upgrade_headers_only_sent_once(self):
...
+
+        # in both cases content-length set so body type is BODY_FLAT. value doesn't matter
+
+        with pytest.raises(TypeError) as exc_info:
+            body = 1234
+            c.request('GET', '/get', body=body, headers={'content-length': str(len(str(body)))})
+        assert str('Request body must be a bytestring, a file-like object returning bytestrings '
+                   'or an iterable of bytestrings. Got: {}'.format(type(body))) in str(exc_info)
+
+        rouge_element = 123
```
Lukasa: Typo: you want `rogue_element` here. =)

Lukasa:
Thanks for this!

I've left a few small notes inline that I'd like you to address before I merge this. =) Additionally, I think switching to use a `TypeError` here, while semantically more accurate, is more churn than I want at this stage in the library's life. Do you mind leaving those as `ValueError`? I'll probably end up wanting to introduce a custom exception for this case anyway.

AvivC:
Done :)

Lukasa:
Fantastic @AvivC! Thanks! :sparkles: :cake: :sparkles:

AvivC:
You're welcome :)

PR #232, opened 2016-04-23
Concurrency update

tbetbetbe:
- syncs the concurrency branch with development branch
- modifies the concurrency implementation to reflect the use of hyper-h2

PR #233, opened 2016-04-24
Concurrency update

tbetbetbe:
- syncs the concurrency branch with development branch
- modifies the concurrency implementation to reflect the use of hyper-h2

Replaces: #232


hyper/http20/connection.py
```diff
@@ -234,9 +301,19 @@ def connect(self):
...
+
+        This method is thread-safe. It may be called from multiple threads, and
+        is a noop for all threads apart from the first.
+
         :returns: Nothing.
+
         """
-        if self._sock is None:
+        with self._lock:
+            if self._sock:
```
Lukasa: Can we keep this as `is None`?

Lukasa: Uh....`is not None`, even?

hyper/http20/connection.py
```diff
@@ -408,32 +510,48 @@ def _new_stream(self, stream_id=None, local_closed=False):
...
-            self._sock.sendall(data)
-        except socket.error as e:
-            if (not tolerate_peer_gone or
-                    e.errno not in (errno.EPIPE, errno.ECONNRESET)):
-                raise
+        # Concurrency
+        #
+        # Synchronizes data writes.  This also ensures that only the current
+        # thread waits for window updates when the flow control window is
+        # blocked.
```
Lukasa: Uh, is this right? No-one is waiting for the window update in this lock.

Lukasa:
@tbetbetbe Thanks for this! I've left some partial notes in the diff, but I think there's a bigger problem we need to address first.

There are a few places where calls into `_conn` are made and followed by calls to `data_to_send()`. This is a bit problematic because it can lead to unsynchronised access to the `_conn` variable (which isn't thread-safe), and can cause the flow of execution to make empty writes (because `data_to_send()` gets emptied by a prior thread). One example of this is in `_adjust_receive_window`, which is as far as I can tell not protected by a lock.

Is this right, or have I missed something?

hyper/http20/connection.py
tbetbetbe: Sorry, that comment was stale - updated

hyper/http20/connection.py
tbetbetbe: Fixed.

tbetbetbe:
You're completely correct - there are a few places where there is unsynchronized access to _conn.  I'll attempt to address these in the next commit

tbetbetbe:
There are travis test failures that I'm unable to reproduce locally at the moment.
I'll look at these again later today, but I hope you don't mind examining the changes until then.

tbetbetbe:
The test failures are consistent:

There is a failure in test_integration because of the check in h2/utilities.py for [pseudo headers](https://github.com/python-hyper/hyper-h2/blob/master/h2/utilities.py#L212)

I think the failure is right.  I'm not sure why it was not failing before - has the h2 dependency been updated recently ?

Lukasa:
Yeah, a new minor release was released today.

tbetbetbe:
ok - travis is passing, this PR now includes the fixes in #236

hyper/http20/connection.py
```diff
@@ -525,7 +668,25 @@ def _recv_cb(self):
...
+        with self._read_lock:
+            log.debug('recv for stream %d with %s already present',
+                      stream_id,
+                      self.recent_recv_streams)
+            if stream_id in self.recent_recv_streams:
+                self.recent_recv_streams.discard(stream_id)
+                return
+
         # TODO: Re-evaluate this.
         self._single_read()
```
Lukasa: I think we have a small locking issue here.

It is technically possible to yield the `_read_lock` from the block above and then be context-switched away before entering `_single_read()`. That defeats the defence given by `recent_recv_streams`. I think `_read_lock` needs to be held across the entirety of this method, with the early-exit being the short-cut most commonly used.

The other option is to do a doubly-checked mechanism where `recent_recv_streams` has its own lock to prevent blocking socket reads from stopping streams that already have data from processing that data, and then making sure we re-check that once this method takes `_read_lock` (in case the previous outstanding read that was blocked on actually returned some data for that stream).

Thoughts?

Lukasa:
Beautiful work @tbetbetbe, I think we're really close here. I have one note about what looks like a hole in the threading, but otherwise I think this is pretty much 100%. If you can address that (either fix it or tell me I'm an idiot, your choice!) then I'll happily go ahead and merge this!

hyper/http20/connection.py
tbetbetbe: Your analysis is spot on, thanks!   I opted for holding _read_lock across the whole method, aiming to keeping the number of locks involved to a minimum.

Lukasa:
So, having gone back and re-reviewed this, I'm still moderately concerned. The `H2Connection` class in `_conn` is not thread-safe *at all*. This means that it basically needs to be guarded by a lock to ensure that two threads don't call into it at the same time for any reason.

This patch doesn't yet guarantee that: as far as I can see there are currently cases where the _conn can be accessed with *only* the read lock or *only* the write lock. That's just not going to be safe at all. =(

Are you up for another round through this patch to resolve that @tbetbetbe?

One possible approach for keeping this somewhat clean is to wrap the `H2Connection` in a custom context manager that basically just acts as a locked container: then you'd do `with self._conn as conn:` everywhere we currently use the connection.

Another option is to write an object that automatically proxies access to the connection with locks. This is tricky, and is *in general* not do-able, but may be do-able in the limited scope of this library.

tbetbetbe:
> Are you up for another round through this patch to resolve that @tbetbetbe?

In for a penny, ...

> One possible approach for keeping this somewhat clean is to wrap the H2Connection in a custom context manager that basically just acts as a locked container: then you'd do with self._conn as conn: everywhere we currently use the connection.

I think I like this; explicit > implicit 👍 
Let me take a look at it.

tbetbetbe:
... in for a pound :smile:

hyper/http20/connection.py
```diff
@@ -34,6 +35,20 @@
...
+    def __init__(self, connection, lock):
+        self.lock = lock
+        self.connection = connection
+
+    def __enter__(self):
+        self.lock.acquire()
+        return self.connection
+
+    def __exit__(self, _exc_type, _exc_val, _exc_tb):
+        self.lock.release()
```
Lukasa: Cool, this class looks good! Let's genericise it a bit though: there's actually nothing specific to the H2Connection about this class, so it can just be `LockedObject`. Additionally, we can move `connection` to `_obj` or something similar, just to indicate that you're really supposed to use the context manager to get it.

hyper/http20/connection.py
```diff
@@ -130,18 +179,26 @@ def __init_state(self):
         users should be strongly discouraged from messing about with connection
         objects themselves.
         """
-        self._conn = h2.connection.H2Connection()
+        self._conn = _LockedH2Connection(h2.connection.H2Connection(),
+                                         self._lock)
```
Lukasa: It feels a little bit weird to use the general lock for this: should we have a lock internal to the locked connection that just serializes there? I'm worried about unnecessary lock contention if we re-use this.

tbetbetbe: Fixed

hyper/http20/connection.py
tbetbetbe: Fixed

Lukasa:
Thanks for that @tbetbetbe, I'll take another review of this tomorrow.

hyper/http20/connection.py
```diff
@@ -260,19 +347,18 @@ def connect(self):
...
-        self._conn.update_settings(
-            {h2.settings.ENABLE_PUSH: int(self._enable_push)}
-        )
-        self._sock.sendall(self._conn.data_to_send())
+        with self._conn as conn:
+            conn.initiate_connection()
+            conn.update_settings(
+                {h2.settings.ENABLE_PUSH: int(self._enable_push)}
+            )
+        self._send_outstanding_data()
```
Lukasa: I think this call may need to be inside the `with` statement here, because `_send_outstanding_data` implicitly uses the connection object.

hyper/http20/connection.py
```diff
@@ -281,31 +367,65 @@ def close(self, error_code=None):
...
+            # Close all streams
+            for stream in list(self.streams.values()):
+                log.debug("Close stream %d" % stream.stream_id)
+                stream.close(error_code)
+
+            # Send GoAway frame to the server
+            try:
+                with self._conn as conn:
+                    conn.close_connection(error_code or 0)
+                self._send_outstanding_data(tolerate_peer_gone=True)
```
Lukasa: This one would also need to be brought into the `with` block.

hyper/http20/connection.py
Lukasa: To be clear, I'm not *certain* of that. It seems to be the case, but I'm willing to be told that I'm wrong and that the `_write` lock protects us. I'm just not sure that it does.

Lukasa: Actually it probably doesn't matter given that `_send_outstanding_data` can tolerate there being no data to send.

Lukasa:
Ok, I think I'm generally happy with this! I'm going to modify a few stylistic things locally, but I expect I'll merge this shortly.

Lukasa:
Merged! :sparkles: :cake: :sparkles: Thanks so much @tbetbetbe!

I'll aim to apply the few remaining changes for this release sometime this week and try to push out a new release with this in it.

tbetbetbe:
great! I am looking forward to the next release 😎

PR #234, opened 2016-04-24
Added missing self.connect() in HTTP11Connection.__enter__

AvivC:
Previously the method was a noop returning self, so
using the with statement on an HTTP11Connection
wouldn't work correctly.
Also added tests for the context manager.

### DIFF
--- a/hyper/http11/connection.py
+++ b/hyper/http11/connection.py
@@ -377,6 +377,7 @@ class HTTP11Connection(object):
     # The following two methods are the implementation of the context manager
     # protocol.
     def __enter__(self):
+        self.connect()
         return self
 
     def __exit__(self, type, value, tb):
--- a/test/test_http11.py
+++ b/test/test_http11.py
@@ -510,6 +510,17 @@ class TestHTTP11Connection(object):
         assert 'File-like bodies must return bytestrings. ' \
                'Got: {}'.format(int) in str(exc_info)
 
+    def test_context_manager_enter_returns_self(self):
+        c = HTTP11Connection('httpbin.org')
+        with c as conn:
+            assert conn is c
+
+    def test_context_manager_connects_and_disconnects(self):
+        c = HTTP11Connection('httpbin.org')
+        with c:
+            assert c._sock is not None
+        assert c._sock is None
+
 
 class TestHTTP11Response(object):
     def test_short_circuit_read(self):


Lukasa:
Thanks!

This shouldn't be needed though: connection is a lazy process that is deferred until we actually need it. Is that insufficient for your case?

AvivC:
I understand, we can just call `request()` and skip calling `connect()` by itself anyway.
I simply fixed the context manager because it was implemented :) If it is unnecessary, IMHO either take it out or merge the fixed version.

AvivC:
Actually now I understand better what you mean.
One can use `with conn_obj ...`, and then the `request()`s inside the block will call `connect()` lazily. So no harm in functionality.

So as I said, I think the lazy nature of things makes the context manager kind of unnecessary.
But if it's there, it should function as expected. A user would expect `with conn_obj` to create a connection, and exiting the block to close it.

AvivC:
Correcting myself: the context manager is not unnecessary because it serves as a `finally` block for the disconnection... Sorry, it's probably the hour here :)

AvivC:
I have removed all my comments. I'm sorry, the time must have got me confused...

The context manager is an easy way for a user to enforce the connection is closed at the end (instead of a `try.. finally`). I now see the reason in having `__enter__` empty.

The question is whether a user would expect `__enter__` to issue the actual connection or not. I'm not sure.

Lukasa:
Hrm, this is complex.

In the first case, I'm not sure I agree. In most cases the expectation around `with` is that it provides deterministic cleanup. It doesn't matter really whether the objects that it needs to clean up are created at the time of entering the block or not: the guarantee of a context manager is that there is no way to *exit* the block without cleanup occurring. That's also why you'd want it here: people generally care *far less*

In this instance, given that connection management is already implicit in the `HTTP20Connection` object (that is, that it expressly does not guarantee to you exactly when the underlying socket connection is opened), I'm not concerned about that implicitness here.

Additionally, it takes almost no work to circumvent the guarantee this patch would add:

```python
c = HTTP20Connection('www.google.com', 443)
c.request('GET', '/')

with c:
    # Do some other stuff
```

Here the connection was created *before* the `with` statement, but will still get cleaned up by the exit of it.

The only purpose for a `with` statement to create the connection, as far as I can see, is that it allows you to deterministically ensure the lifetime of a connection. I just don't believe that's as big a gain as the loss we'd encounter by eagerly creating the connection, rather than lazily doing it.

AvivC:
I tend to agree :) Closing the PR.

Lukasa:
Thanks for the contribution nonetheless!

AvivC:
Sure :) will keep proposing patches if I come across ideas.

Lukasa:
Please do! Ideas are 100x better if they come with patches. =D

PR #235, opened 2016-04-25
Raising error in HTTP11Response if created with illegal attributes

AvivC:
An HTTP response must either specify a content-length header, specify
'close' for a connection header to signal that the connection will be
closed after the response, or be a chunked response.

If none of these conditions are true, we raise a detailed ValueError,
instead of the plain assertion.


Lukasa:
Thanks for this! The patch fails the linter: do you mind cleaning that up?

PR #236, opened 2016-04-26
Correct a test that fails sporadically

tbetbetbe:
- this is really just replacing str with bytes.
- the body is initialized outside the test to make it easy to tell what actually failed

### DIFF
--- a/test/test_http11.py
+++ b/test/test_http11.py
@@ -477,15 +477,16 @@ class TestHTTP11Connection(object):
         c = HTTP11Connection('httpbin.org')
 
         rogue_element = 123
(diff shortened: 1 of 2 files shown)


Lukasa:
While we're fixing tests, do you want to fixup the other failing test? `test_receiving_trailers` and `test_receiving_trailers_before_reading` both use a custom pseudo-header field, which RFC 7540 forbids. Mind changing it to something else?

tbetbetbe:
ok  - will do shortly.

tbetbetbe:
Done

Lukasa:
Fab, thanks @tbetbetbe! :sparkles: :cake: :sparkles:

PR #243, opened 2016-05-24
pants/pex cannot handle compatible release versioning; rewrite hyperf…

agandhi4:
…rame dependency

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -53,7 +53,7 @@ except AttributeError:
 
 
 def resolve_install_requires():
-    basic_dependencies = ['h2>=2.3,<3.0', 'hyperframe~=3.2']
+    basic_dependencies = ['h2>=2.3,<3.0', 'hyperframe>=3.2,<4.0']
 
     if py_version == (3, 3):
         basic_dependencies.extend(


Lukasa:
So, I'm open to making this change, but I'm only going to do it if I can see any movement at all from pex to support this versioning scheme. [PEP 440](https://www.python.org/dev/peps/pep-0440/) has defined `~=` as part of a valid version numbering scheme since 2014, so it's not like this is a *surprise* to anyone. Can I get a link to an issue tracking this in Pex, please?

sigmavirus24:
@Lukasa looks like it's https://github.com/pantsbuild/pex/issues/50

Lukasa:
I think pantsbuild/pex#266 opened by @agandhi4 is the better example: pantsbuild/pex#50 has remained open for a year, so another prod seems reasonable.

Regardless, now that we have an issue to track, I'll merge this. Thanks @agandhi4 for working around this and for trying to improve these projects. =)

agandhi4:
@Lukasa Thanks! I love the library, and hopefully will be able to contribute more as I use this more 👍

Issue #250, opened 2016-05-31
Proxied call to `fileno` fails for `hyper.ssl_compat.SSLSocket`


benlast:
The class `hyper.ssl_compat.SSLSocket` uses this trick to reduce boilerplate:
```python
    # a dash of magic to reduce boilerplate
    methods = ['accept', 'bind', 'close', 'getsockname', 'listen', 'fileno']
    for method in methods:
        locals()[method] = _proxy(method)
```
where:

 ```python
def _proxy(method):
    def inner(self, *args, **kwargs):
        getattr(self._conn, method)(*args, **kwargs)
    return inner
```

This does not work in at least one case: calling `fileno` on the `SSLSocket` returns `None` where the underlying wrapped socket's `fileno` returns an integer value. This causes the check `while count and self._sock is not None and self._sock.can_read` (`hyper.http20.connection.py` line 748, in `hyper.http20.connection.HTTP20Connection._recv_cb`) to fail (raises an exception).

I believe that `_proxy` should be defined as:

```python
def _proxy(method):
    def inner(self, *args, **kwargs):
        return getattr(self._conn, method)(*args, **kwargs)
    return inner
```
Note the extra `return` in `inner`

PR #251, opened 2016-05-31
Add missing `return` to `_proxy` on `SSLSocket`

benlast:
See the issue at https://github.com/Lukasa/hyper/issues/250


Lukasa:
Thanks so much! :sparkles: :cake: :sparkles:

PR #252, opened 2016-06-02
Include ssl dependencies into ssl_compat for python versions <2.7.9

agandhi4:
I'm not sure what the long term plan is for ssl_compat, but this is what I'm using on my machine as I am pinned to a version of python lower than 2.7.9. If there is a better way, let me know!

### DIFF
--- a/hyper/ssl_compat.py
+++ b/hyper/ssl_compat.py
@@ -48,7 +48,8 @@ def _proxy(method):
         return getattr(self._conn, method)(*args, **kwargs)
     return inner
 
-
+# Referenced in hyper/http20/connection.py. These values come from the python ssl package, and
+# must be defined in this file for hyper to work in python versions <2.7.9
 SSL_ERROR_WANT_READ = 2
 SSL_ERROR_WANT_WRITE = 3
 
--- a/hyper/ssl_compat.py
+++ b/hyper/ssl_compat.py
@@ -48,8 +48,9 @@ def _proxy(method):
         return getattr(self._conn, method)(*args, **kwargs)
     return inner
 
-# Referenced in hyper/http20/connection.py. These values come from the python ssl package, and
-# must be defined in this file for hyper to work in python versions <2.7.9
+# Referenced in hyper/http20/connection.py. These values come 
+# from the python ssl package, and must be defined in this file 
+# for hyper to work in python versions <2.7.9
 SSL_ERROR_WANT_READ = 2
 SSL_ERROR_WANT_WRITE = 3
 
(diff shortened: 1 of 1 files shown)


agandhi4:
Actually this still doesn't fix all my issues, I have this issue with python 2.7.7:

```
hyper/http20/connection.py:35: in <module>
                         TRANSIENT_SSL_ERRORS = (ssl.SSL_ERROR_WANT_READ, ssl.SSL_ERROR_WANT_WRITE)
                     E   AttributeError: 'NoneType' object has no attribute 'SSL_ERROR_WANT_READ'
```

Why can't we just import ssl for all versions of python==2.7.x?

Lukasa:
Python 2.7.X where X < 9 does not expose an SSLContext object that is of any value. In particular, we cannot negotiate ALPN/NPN, we cannot disable TLS compression, and generally speaking are unable to configure any TLS configuration that matters.

agandhi4:
@Lukasa 

This PR fixes the issue described, in that we don't get these values from the enum in _ssl.c. However, there is another issue that we should look at - namely that when you install the package on python < 2.7.9, the service-identity package is not being included. You wrap the import error in a try/except, so you don't see the error, but the solution is to just install the package separately.

I don't know enough about how packages are bundled in python to know how to fix the issue. Why can't we just make it a requirement for all versions?

Lukasa:
There's no point using `service-identity` with 2.7.9 or later, because they include their own hostname verification functions. Again, `service-identity` should be being installed automatically on those versions of Python where PyOpenSSL is installed.

agandhi4:
I'm not sure what you want the behavior to be, but I'll create an issue. Here are some very basic steps:

1) On a clean system of python 2.7.9 or earlier, pip install hyper.
2) Open up python REPL.
3) `import hyper`
4) Get the following error `AttributeError: 'NoneType' object has no attribute 'SSL_ERROR_WANT_READ'`

This is because pyOpenSSL and service-identity are not being installed. Only reason I found that out was I commented out the code that swallows the import error when you import ssl_compat. What do you want the behavior to be? Should the users of the package know to install pyOpenSSL and service-identity manually?

Lukasa:
Argh, I know what's going on here. While `setup.py` is conditionally adding the relevant dependencies, that doesn't work for a wheel as installing a wheel doesn't execute code. We'll need to add the conditional dependencies declaratively, not programmatically.

Lukasa:
This should have been resolved in #253, and will take effect in the next release.

agandhi4:
Hey @Lukasa,

We still need the values for SSL_ERROR_WANT_READ and SSL_ERROR_WANT_WRITE in ssl_compat.py.

I finally am using a version of python greater than 2.7.9, but this should be put in anyways.

Lukasa:
Hrm, good spot.

Lukasa:
I've had to clean up some of the syntax stuff: see #259.

PR #253, opened 2016-06-07
Use markers for conditional dependencies.

Lukasa:
Resolves #252.

/cc @dstufft

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -89,7 +89,7 @@ setup(
     extras_require={
         'fast': ['pycohttpparser'],
         # Fallback to good SSL on bad Python versions.
-        ':python_full_version < "2.7.9" or python_version == "3.3"': [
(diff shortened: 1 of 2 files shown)


dstufft:
🚢

setup.py
```diff
@@ -114,5 +88,14 @@ def resolve_install_requires():
     },
     extras_require={
         'fast': ['pycohttpparser'],
+        # Fallback to good SSL on bad Python versions.
+        ':python_full_version < "2.7.9" or python_version == "3.3"': [
```
sigmavirus24: You don't indicate that 3.3 is supported in the classifiers. Why are you accommodating it here then?

Lukasa: That's an excellent question.

Lukasa:
Cool, this should be fine now.

Issue #246, opened 2016-05-28
Unsynchronized access to connection object from streams.


Lukasa:
During stream creation, streams are provided access to the underlying connection object without its lock wrapper. This allows streams to perform unsynchronised access to the connection object, which is BAD.

While we’re here we should evaluate whether streams should be using _send_outstanding_data instead of the _send_cb, and indeed whether they should be just handed a reference to the connection instead of the various CBs.

fredthomsen:
Seems that passing in the LockedObject and not just the underlying `H2Connection` where Stream is constructed in the `HTTP20Connection` solves the first problem.  Perhaps I am not completely understanding and you can point out the error of my ways.

As far as the passing in a reference, I assume you mean to the `HTTP20Connection`.  That seems very much like a layering/design pattern violation no?  Is there some reason why the callbacks are frowned upon?

Lukasa:
@fredthomsen You're entirely right about the first problem, that would solve it. You'd also need to rewrite a lot of code to use `with` statements to grab the lock, but that's ok.

As to passing in a reference, there's no way around it: there's a layering violation going to happen here somewhere because connection objects call streams and then streams call connection objects. The only way *around* that is to heavily refactor the code so that streams no longer need to call the connection *at all*, which is certainly do-able. Otherwise, one way or another, we need to pass *something* to the stream. The big problem right now is that the stream doesn't have access to `send_outstanding_data`, which I'd much rather they used than the current `send_cb` that they do use.

fredthomsen:
This feel icky.  Made a PR for now.

PR #254, opened 2016-06-08
Synchronize access to h2connection from stream

fredthomsen:
This is the simple fix with layering violations.  Not sure how I feel about this, and not done yet. Meant to handle issue #246.

### DIFF
--- a/hyper/http20/stream.py
+++ b/hyper/http20/stream.py
@@ -303,7 +303,9 @@ class Stream(object):
         The size of our outbound flow control window.
         """
 
-        return self._conn._obj.local_flow_control_window(self.stream_id)
+        with self._conn as conn:
+            out_fc_with_size = conn.local_flow_control_window(self.stream_id)
+        return out_fc_with_size
 
     def _send_chunk(self, data, final):
         """
--- a/hyper/http20/stream.py
+++ b/hyper/http20/stream.py
@@ -96,7 +96,7 @@ class Stream(object):
         headers = self.get_headers()
         with self._conn as conn:
             conn.send_headers(self.stream_id, headers, end_stream)
-            self._send_cb(conn.data_to_send())
+        self._send_cb()
 
         if end_stream:
             self.local_closed = True
@@ -191,7 +191,7 @@ class Stream(object):
                 conn.increment_flow_control_window(
                     increment, stream_id=self.stream_id
                 )
-                self._send_cb(conn.data_to_send())
+            self._send_cb()
 
     def receive_end_stream(self, event):
         """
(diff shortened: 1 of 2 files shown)


hyper/http20/stream.py
```diff
@@ -297,7 +302,8 @@ def _out_flow_control_window(self):
         """
         The size of our outbound flow control window.
         """
-        return self._conn.local_flow_control_window(self.stream_id)
+
+        return self._conn._obj.local_flow_control_window(self.stream_id)
```
Lukasa: This needs to acquire the lock.

hyper/http20/stream.py
```diff
@@ -94,8 +94,9 @@ def send_headers(self, end_stream=False):
         Sends the complete saved header block on the stream.
         """
         headers = self.get_headers()
-        self._conn.send_headers(self.stream_id, headers, end_stream)
-        self._send_cb(self._conn.data_to_send())
+        with self._conn as conn:
+            conn.send_headers(self.stream_id, headers, end_stream)
+        self._send_cb()
```
Lukasa: Should we maybe rename the `_send_cb`? I'd like it to match the name as used in the connection object if possible.

Lukasa:
I'm currently super-impressed that this didn't fail a whole bunch of tests: I'd have expected it to. Regardless, this seems like a good start.

I, like you, feel icky about this, but so far it's the best we've got. We definitely need a refactor down the road.

hyper/http20/stream.py
fredthomsen: Serious lack of coffee here...

hyper/http20/stream.py
fredthomsen: Understood.  I didn't hold a strong opinion on this one way or another.

fredthomsen:
Ok updates made.  Should another issue be opened for refactoring the connection stream interaction/layering refactoring?  Any concerns about additional tests needed here?

hyper/http20/stream.py
```diff
@@ -277,16 +279,19 @@ def close(self, error_code=None):
...
+                try:
+                    conn.reset_stream(self.stream_id, error_code or 0)
+                except h2.exceptions.ProtocolError:
+                    # If for any reason we can't reset the stream, just
+                    # tolerate it.
+                    pass
+                else:
+                    send = True
+            if send:
+                self._send_outstanding_data(tolerate_peer_gone=True)
```
Lukasa: You may as well just bring this directly into the `else`, rather than test the conditional.

hyper/http20/stream.py
```diff
@@ -297,7 +302,10 @@ def _out_flow_control_window(self):
         """
         The size of our outbound flow control window.
         """
-        return self._conn.local_flow_control_window(self.stream_id)
+
+        with self._conn as conn:
+            out_fc_with_size = conn.local_flow_control_window(self.stream_id)
+        return out_fc_with_size
```
Lukasa: You can put the `return` in the `with` block, Python will do the right thing.

Lukasa:
Cool, just a few style nits and I'll be happy.

As to the layering violation, yeah, I think I'd like a new issue for that. It'll require someone to sit down and think real hard about the code layout for a bit, but we'll see how that goes.

fredthomsen:
Alright, updates made.

Lukasa:
Thanks @fredthomsen! Made a minor change myself. =)

PR #255, opened 2016-06-09
Issue 249: Don't rely on the bundled certificate file being available.

JasonGowthorpe:


hyper/tls.py
```diff
@@ -29,14 +29,16 @@ def wrap_socket(sock, server_hostname, ssl_context=None, force_proto=None):
     A vastly simplified SSL wrapping function. We'll probably extend this to
     do more things later.
     """
-    global _context
+    if ssl_context:
+        # if an SSLContext is provided then use it instead of default context
+        _ssl_context = ssl_context
+    else:
+        global _context
```
Lukasa: I...don't think you can do this. `global _context`, however, is just a statement that makes it clear to Python that I want to be able to assign to the global variable, so it shouldn't be needed.

Lukasa: Or, more specifically, it should be hoisted to the top of the function block.

hyper/tls.py
```diff
@@ -94,9 +96,17 @@ def init_context(cert_path=None, cert=None, cert_password=None):
         encrypted and no password is needed.
     :returns: An ``SSLContext`` correctly set up for HTTP/2.
     """
+    cafile = cert_path or cert_loc
+    if not cafile or not path.exists(cafile):
+        errMsg = ("No certificate found at " + str(cafile) + ". Either " +
+                  "ensure the default cert.pem file is included in the " +
+                  "distribution or provide a custom certificate when " +
+                  "creating the connection.")
+        raise Exception(errMsg)
```
Lukasa: Please don't raise plain `Exception`s: define a custom exception in our `exceptions.py` file and raise that. =)

Lukasa:
Thanks for this! I have a few small notes, and then we need a test to deal with the new code branch you've introduced. =) Are you happy to deal with that?

JasonGowthorpe:
Making the suggested changes now.
Out of interest, what is the issue with only declaring the global _context in the necessary code path?
Is it just a stylistic point or something that's changed in newer versions of Python (I have never used 3+)?
Where would you like a test adding, just in test_hyper.py?

Lukasa:
It's stylistic. =)

As for tests, maybe in `test_SSLContext.py`?

test/test_SSLContext.py
```diff
@@ -60,3 +61,51 @@ def test_client_certificates(self):
...
+
+        conn = hyper.HTTP20Connection('http2bin.org', 443, ssl_context=context)
+        conn.request('GET', '/', None, {})
+
+        hyper.tls.cert_loc = backup_cert_loc
+        hyper.tls._context = backup_context
+
+        assert conn.ssl_context.check_hostname
+        assert conn.ssl_context.verify_mode == ssl.CERT_REQUIRED
+        assert conn.ssl_context.options & ssl.OP_NO_COMPRESSION != 0
```
Lukasa: In this instance, while I like the extremely thorough testing you've done here, I think we can safely just test `_init_context` directly. That's the only place where we're concerned.

If you want a *slightly* more thorough test, you can actually wrap a socket but not bother to connect it. In both of these cases we drastically reduce the amount of system access needed to test the software, which is good.

Lukasa:
Thanks for the work! I have one tiny style note and one thing about the testing. As always, I appreciate your spending time here, and if you get sick of the code review just say so and I can make whatever changes are still required myself.

JasonGowthorpe:
I'll get there eventually, hopefully it's up to your exacting standards now :)

Lukasa:
Cool, this looks good to me. I'm not going to merge it just this second because I want to be in front of a laptop so I can make some associated docs changes.. But this looks really good, thank you so much for the work! :sparkles: :cake: :sparkles:

Lukasa:
Thanks @JasonGowthorpe! :sparkles: :cake: :sparkles:

Issue #256, opened 2016-06-29
 self._sock is None inside of _single_read


72squared:
I opened this with the apns2 package maintainer since that is where I discovered the bug but I think this is really a problem inside of hyper:

https://github.com/Pr0Ger/PyAPNs2/issues/11

 self._sock is None when it hits this line:

https://github.com/Lukasa/hyper/blob/development/hyper/http20/connection.py#L644

It is a rare occurance but it does happen. Any thoughts on what could be causing it?

Lukasa:
So the `_sock` is set to `None` if the connection is by the remote peer, either cleanly or not.

This is definitely a behaviour I don't love: if the remote peer closes the connection with error code NO_ERROR, hyper doesn't handle it very gracefully. The ideal behaviour would be to quiesce the connection. That's quite a substantial change, however.

In the meantime, can I get a feel for how you're using the code? Are you running multithreaded, or from a single thread?

Lukasa:
In the short term the easiest fix is probably to raise an exception in all connection termination cases: at least that way your code would be able to spot this and handle it as appropriate.

72squared:
I am using the library through pyapns2:

https://github.com/Pr0Ger/PyAPNs2

We schedule an apple push notification to an iOS device as a background job, and a thread from a thread pool in a background job handles the request and uses pyapns2 (which uses hyper) to send a push notification through apple's http2 api backend.

I am going to try to see if it makes a difference if I use Greenlet thread workers rather than python native threads. But because I see it happen simultaneously on different servers I suspect it is apple's API dropping connections. 

In that case, an exception on connection termination would be fine by me because it would be much easier for my application to correctly handle. A TypeError exception is hard to do anything intelligent about.

72squared:
Could you make it raise a Connection exception if the connection closes and there are still open stream requests associated with the connection? Not sure if that is possible. Because I think if the connection gets closed and there are no outstanding requests there is no real problem. The next request would cause a new connection to be made. It's only if you are waiting on a response or in the middle of sending a request that there is a problem.

72squared:
Looking at the code, here's where I can see a problem:

https://github.com/Lukasa/hyper/blob/development/hyper/http20/connection.py#L678-L682

If you have multiple streams outstanding and the connection terminates, you close the connection, making self._sock = None. then if other threads are in progress reading from different streams on the same connection at the same time they could run into a problem in lots of other spots.

Lukasa:
[...]

PR #257, opened 2016-06-30
potential fix for multiple stream reads when sock closes

72squared:
to address issue #256

### DIFF
--- a/hyper/http20/connection.py
+++ b/hyper/http20/connection.py
@@ -641,6 +641,8 @@ class HTTP20Connection(object):
         #
         # I/O occurs while the lock is held; waiting threads will see a delay.
         with self._read_lock:
+            if self._sock is None:
+                raise ConnectionError('tried to read after connection close')
             self._sock.fill()
             data = self._sock.buffer.tobytes()
             self._sock.advance_buffer(len(data))
@@ -739,6 +741,12 @@ class HTTP20Connection(object):
             if stream_id in self.recent_recv_streams:
                 self.recent_recv_streams.discard(stream_id)
                 return
+
+            # make sure to validate the stream is readable.
+            # if the connection was reset, this stream id won't appear in
+            # self.streams and will cause this call to raise an exception.
+            if stream_id:
+                self._get_stream(stream_id)
 
             # TODO: Re-evaluate this.
             self._single_read()
--- a/test/test_hyper.py
+++ b/test/test_hyper.py
@@ -8,7 +8,7 @@ from hyperframe.frame import (
 from hpack.hpack_compat import Encoder
 from hyper.http20.connection import HTTP20Connection
 from hyper.http20.response import HTTP20Response, HTTP20Push
-from hyper.http20.exceptions import ConnectionError
+from hyper.http20.exceptions import ConnectionError, StreamResetError
 from hyper.http20.util import (
     combine_repeated_headers, split_repeated_headers, h2_safe_headers
 )
@@ -22,7 +22,6 @@ import pytest
 import socket
 import zlib
 from io import BytesIO
-
 
 TEST_DIR = os.path.abspath(os.path.dirname(__file__))
 TEST_CERTS_DIR = os.path.join(TEST_DIR, 'certs')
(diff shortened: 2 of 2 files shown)


72squared:
cleaned up my earlier lint errors and simplified the code since there was a case where I realized my patch was doing some redundant logic that already existed inside close. All tests pass now, yay! Look it over and let me know what you think.

hyper/http20/connection.py
```diff
@@ -740,6 +742,12 @@ def _recv_cb(self, stream_id=0):
                 self.recent_recv_streams.discard(stream_id)
                 return
 
+            # make sure to validate the stream is readable.
+            # if the connection was reset, this stream id won't appear in
+            # self.streams and will cause this call to raise an exception.
+            if stream_id:
```
Lukasa: This check isn't right: I think you want `if stream_id in self.streams`.

Lukasa: Oh, I'm sorry, I totally misread this code. This is fine. =)

Lukasa:
So I really like this, thanks so much!

The only change I think I'd like is to have a *more specific* form of `ConnectionError` that we throw in the case where we read from a `None` socket. That would allow us to correctly spot this behaviour in code to understand the difference between the case where a general connection problem occurred and a case where the connection has been torn down by the remote peer. `ConnectionTerminatedError`, perhaps?

I also want to think a bit about whether we can do better in the case of a stream attempting to read and discovering that it is no longer present other than `StreamResetError`, but probably we can't. =)

Also, feel free to add yourself to CONTRIBUTORS.rst. =)

This is a model patch: despite having a fairly complex problem to solve, it solves it relatively well. I hope we can make this even better in future: I'm considering refactoring a lot of this code because it's vastly more complex than it needs to be.

hyper/http20/connection.py
72squared: one way I could make it more clear is something like:

```
if stream_id != 0:
```

What kind of values do we see as stream ids? Are they sequential integers always? This could be a problem with my other check where I bail if the stream_id isnt in self.streams because the new socket connection could generate another stream id that is another integer that is identical to the one we currently have. I need to read the http2 spec.

72squared:
One idea I experimented with and left out of the patch was passing the event `h2.events.ConnectionTerminated` to `Stream.receive_reset(event)` on this line:

https://github.com/Lukasa/hyper/blob/development/hyper/http20/connection.py#L681

so that it looked something like:

```
map(lambda s: s.receive_reset(event), self.streams.values())
```

And then attach that event to a stream property like `self.remote_close_event` so that we could inspect that value and throw a more appropriate error. But in the end I left it out. There's something to be said for keeping a patch really simple and to the point. Solve it with the most minimal changeset first and then improve after that.

72squared:
As far as whether or not to use ConnectionTerminatedError vs ConnectionError for when self._sock is None, I think that complicates things unecessarily. We have the backtrace in case we are looking. The comment on ConnectionError says:

```
   The remote party signalled an error affecting the entire HTTP/2
    connection, and the connection has been closed.
```

Which matches the case well enough. If you feel strongly about it I suppose we could extend ConnectionError with  ConnectionTerminatedError so you can just look for ConnectionError or look for ConnectionTerminatedError if you really want to. But I really can't see application developers writing code to use it that way so it seems pointless to me.

hyper/http20/connection.py
Lukasa: Stream IDs *are* sequential. This is part of why I was wondering whether it is really acceptable at all to re-use a connection object in this manner.

Lukasa:
Yeah, I suppose that argument about `ConnectionError` makes some sense. =) In that case, let's merge this for now: it's a substantial improvement over the status quo. As you've pointed out, there are further follow-on questions, but this takes us to a much better place.

Lukasa:
Thanks @72squared! :sparkles: :cake: :sparkles:

72squared:
You are welcome. One thought about differentiating on sequential stream ids and multiple connections, you could always return a stream_id that is a tuple of the stream_id and the current socket connection or some other random identifier that is a lookup to the stream for the current sock. It adds a layer of indirection to your lookup which is a pain in the arse but it would prevent problems of sequential stream ids from different socket connections colliding.

Lukasa:
Frankly, I feel like when we're bending over backwards that much to get this to function it's a sign of a bad design. =P

72squared:
Yeah some better encapsulation of the sock and how it matches up with the stream id seems more appropriate.

72squared:
As far as being added to Contributors, I don't care. You can add me if you want when you cut the next version release but it's not important to me. I just like bugs fixed so my software works as expected.

:-)

Issue #265, opened 2016-08-04
Stream is not removed from recent_recv_streams when cleanly closed.


Lukasa:
When `_send_rst_frame` is called we evict streams from `recent_recv_streams`, but we don't do it when `_stream_close_cb` is called. That seems wrong.

PR #270, opened 2016-08-16
removing stream from recent_recv_streams on clean close

nateprewitt:
This addresses #265. Not sure if the test is overkill for such a minor change, but I figured I'd err on the side of 101% coverage :) I'll strip it out if it's unneeded.

### DIFF
--- a/hyper/http20/connection.py
+++ b/hyper/http20/connection.py
@@ -817,6 +817,7 @@ class HTTP20Connection(object):
         """
         try:
             del self.streams[stream_id]
+            self.recent_recv_streams.discard(stream_id)
         except KeyError:
             pass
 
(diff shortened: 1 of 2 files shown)


Lukasa:
Thanks for this!

I'm happy with the change, and have no objection to that test, but I'd also like a more specific test. Can we write a test that cleanly closes one stream? That is, that uses a stream through its lifetime, such that it gets a complete response, and then confirms that it has been removed from everything?

nateprewitt:
Were you looking for something more along the lines of [this](https://github.com/nateprewitt/hyper/commit/624afca#diff-3f0c2e4380c8aaed128603ed40d5f22fR262)? This is what I originally wrote but felt it might exercise unnecessary code.

Lukasa:
Yup @nateprewitt, that looks right.

nateprewitt:
Replaced old test with lifecycle test. The former felt redundant with this one.

Lukasa:
Fab, thanks @nateprewitt! :sparkles: :cake: :sparkles:

PR #273, opened 2016-08-18
Send ping frame

plucury:
I'm using hyper to send messages to APNS, and it suggests checking the health of the connection with ping frame. However, I think it should be useful in other use cases.

### DIFF
--- a/test/test_hyper.py
+++ b/test/test_hyper.py
@@ -5,7 +5,7 @@ from h2.frame_buffer import FrameBuffer
 from hyperframe.frame import (
     Frame, DataFrame, RstStreamFrame, SettingsFrame, PushPromiseFrame,
     WindowUpdateFrame, HeadersFrame, ContinuationFrame, GoAwayFrame,
-    FRAME_MAX_ALLOWED_LEN
+    PingFrame, FRAME_MAX_ALLOWED_LEN
 )
 from hpack.hpack_compat import Encoder
 from hyper.http20.connection import HTTP20Connection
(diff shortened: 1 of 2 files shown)


test/test_hyper.py
```diff
@@ -80,6 +80,20 @@ def test_connections_can_parse_ipv6_hosts_and_ports(self):
...
+
+        c = HTTP20Connection('www.google.com')
+        c._sock = DummySocket()
+        c._send_cb = data_callback
+        c.ping('00000000')
+
+        frames = list(frame_buffer)
+        assert len(frames) == 1
+        f = frames[0]
+        assert isinstance(f, PingFrame)
```
Lukasa: Can we enhance this check to confirm that the opaque data is passed through?

Lukasa:
Thanks for this! This change generally looks good to me, though I have left one inline note.

plucury:
@Lukasa Thanks!  Just updated my code.

Lukasa:
Fab! I'm currently sat on a plane which means I can't effectively do the extra steps I like to do when I merge a PR, so if you don't mind I'll merge this a bit later in the day when I'm no longer awkwardly sat on an aircraft (or this flight might have wifi, in which case I'll do it here when this plane eventually takes off!)

plucury:
@Lukasa That's fine. Have a nice trip!

Lukasa:
Thanks @plucury! :sparkles: :cake: :sparkles:

PR #276, opened 2016-08-19
Reset connection state when it was closed even its _sock object is None 

plucury:
So I can reuse that connection : )

### DIFF
--- a/test/test_hyper.py
+++ b/test/test_hyper.py
@@ -2,6 +2,7 @@
 import h2.settings
 
 from h2.frame_buffer import FrameBuffer
+from h2.connection import ConnectionState
 from hyperframe.frame import (
     Frame, DataFrame, RstStreamFrame, SettingsFrame, PushPromiseFrame,
     WindowUpdateFrame, HeadersFrame, ContinuationFrame, GoAwayFrame,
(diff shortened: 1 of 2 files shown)


test/test_hyper.py
```diff
@@ -275,6 +276,18 @@ def test_closed_connections_are_reset(self):
         assert c.recent_stream is None
         assert c.next_stream_id == 1
         assert c.window_manager is not wm
+        with c._conn as conn:
+            assert conn.state_machine.state == ConnectionState.IDLE
+
+        c = HTTP20Connection('www.google.com')
```
Lukasa: Can we make a positive assertion of what the state of this object is here?

Lukasa:
Thanks @plucury! This looks great, I have one small note about the tests.

test/test_hyper.py
plucury: @Lukasa Maybe we can reuse the connection above since it was already reseted.  And then asserting its state when it was closed again.

Lukasa: What I'm getting at is, assuming this patch changes something in the code, there should be some difference in the state of the connection between this point and the point after `close`. If there isn't, we're not testing the correct behaviour. ;)

plucury: @Lukasa I'm a little confused. The state of connection before or after `close()` should both be `IDLE`. But it holds the different `H2ConnectionStateMachine` instances.  So should I check the state of the origin state machine or manually change its state here?

Lukasa: @plucury Let me rephrase.

The purpose of this change is to enable some specific behaviour, yes? That means that, before this change, there was presumably some resetting that was not happening, and after this change there is some resetting that *is* happening. We should write a test that confirms this behaviour: namely, that test should fail *before* this patch applies and pass afterward.

Is that true of the test written here? If it isn't, we should rewrite it so that it is.

plucury: @Lukasa Yes, if we close a connection without `_sock` object before this patch. The state of that connection should be `CLOSED` and it would make assertion `conn.state_machine.state == ConnectionState.IDLE`  fail. On the other hand,  maybe we can check if the connection holds a different `H2Connection` instance to make it more clearly.

Lukasa:
Cool, I'm happy with this change. Thanks for working with me @plucury! :sparkles:

plucury:
@Lukasa Thanks for your patience too. 🍻 🍻

PR #277, opened 2016-08-24
added socket_timeout to HTTPConnection

johnwheeler:
Hello Cory,

This is my first attempt. Please let me know your thoughts. 

Thank you

### DIFF
--- a/hyper/http11/connection.py
+++ b/hyper/http11/connection.py
@@ -55,9 +55,13 @@ class HTTP11Connection(object):
     :param proxy_port: (optional) The proxy port to connect to. If not provided
         and one also isn't provided in the ``proxy`` parameter,
         defaults to 8080.
+    :param socket_timeout: (optional) Specify a timeout for the underlying
+        socket. Defaults to ``None``. For more information on socket timeouts,
+        see :meth:`socket.socket.settimeout`.
     """
     def __init__(self, host, port=None, secure=None, ssl_context=None,
-                 proxy_host=None, proxy_port=None, **kwargs):
+                 proxy_host=None, proxy_port=None, socket_timeout=None,
+                 **kwargs):
         if port is None:
             self.host, self.port = to_host_port_tuple(host, default_port=80)
         else:
@@ -78,6 +82,7 @@ class HTTP11Connection(object):
 
         self.ssl_context = ssl_context
         self._sock = None
+        self._socket_timeout = socket_timeout
 
         # Setup proxy details if applicable.
         if proxy_host:
@@ -117,6 +122,7 @@ class HTTP11Connection(object):
                 port = self.proxy_port
 
             sock = socket.create_connection((host, port), 5)
+            sock.settimeout(self._socket_timeout)
             proto = None
 
             if self.secure:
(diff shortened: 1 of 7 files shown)


hyper/http11/connection.py
```diff
@@ -117,6 +122,7 @@ def connect(self):
                 port = self.proxy_port
 
             sock = socket.create_connection((host, port), 5)
```
Lukasa: We probably want to use the timeout in `create_connection` too: see [the docs](https://docs.python.org/2/library/socket.html#socket.create_connection) for how to do it.

hyper/http20/connection.py
```diff
@@ -361,6 +366,7 @@ def connect(self):
                 port = self.proxy_port
 
             sock = socket.create_connection((host, port))
```
Lukasa: Same note as above.

test/test_abstraction.py
```diff
@@ -10,22 +10,23 @@ def test_h1_kwargs(self):
...
+            'socket_timeout': None,
             'other_kwarg': True,
         }
 
     def test_h2_kwargs(self):
         c = HTTPConnection(
             'test', 443, secure=False, window_manager=True, enable_push=True,
             ssl_context=True, proxy_host=False, proxy_port=False,
-            other_kwarg=True
+            socket_timeout=None, other_kwarg=True
```
Lukasa: Ideally we want to set this to a non-None value to confirm it's not getting defaulted. Let's try....`10`.

test/test_http11.py
```diff
@@ -511,6 +512,13 @@ def read(self, size):
         assert 'File-like bodies must return bytestrings. ' \
                'Got: {}'.format(int) in str(exc_info)
 
+    def test_exception_raised_on_socket_timeout(self):
+        c = HTTP11Connection('httpbin.org', socket_timeout=0.01)
```
Lukasa: Generally it's good to set this value comedically small so that running on faster machines doesn't accidentally get this to work.

Lukasa:
Thanks for this! In the first instance I've provided a few specific notes in the diff. Please take a look at those and see if you can apply them. Let me know if you need anything extra!

test/test_http11.py
johnwheeler: Thank you for taking the time to make these comments. I will adjust the changes above. For this line comment, the URL I am requesting is designed to block for n seconds

 c.request('GET', '/delay/1')

From the docs on httpbin "Delays responding for min(n, 10) seconds."

so I think that would save us from even Moore's Law. Please advise if I'm overlooking something.

Lukasa: Ah, yes, sorry, you're right.

johnwheeler:
I made the changes. I removed the test case I added because since `socket_timeout` is used for both read and connect timeout, the read timeout can't be exercised (no read if no connect). this can be solved by having separate timeout values. If that's your preference, please let me know. I am indifferent at this time.

johnwheeler:
Ugh - I'm sorry for being wishy-washy. I was looking at how requests does it

http://docs.python-requests.org/en/master/user/advanced/ Timeout section

Maybe we should have separate read and connect timeouts with this tuple based approach? 

I'm in no hurry to get this merged and want to do it the right way, so please take your time to think about this, and I'll do the work.

Lukasa:
I'm open to having separate timeout values, sure. =)

Issue #275, opened 2016-08-19
Possible deadlock with HTTP20Connection


plucury:
`HTTP20Connection.connect()` acquires `_lock` when it was started, and try to acquire `_write_lock` in `_send_preamble()`. However `HTTP20Connection.request()` acquires `_write_lock` first, then it will try to acquire `_lock` in method `_new_stream()` and `connect()` . This should cause deadlock if there are two threads call them at the same time.

I'm not sure if it is possible to call `connect()` and `_new_stream()` before acquire `_write_lock` in `connect()` to avoid it.

Here is a simple script will stuck on deadlock: https://gist.github.com/plucury/f380d191ca7cc9f920e3f6d09dd5fefa

Lukasa:
Hrm.

That definitely seems like a problem. Based on the rules documented in the comment, that means that `request()` needs to acquire `_lock` *first*, before it acquires `_write_lock`. I think the way we want to do that is to hoist `connect()` calls out if we can.

Probably the way we need to deal with this is that we need to stop using `putrequest` and `endheaders` directly in the code. Likely we need to turn those into wrapper methods that grab the correct locks, and then have internal methods that don't grab the locks at all so that we can manage them ourselves in `request`.

plucury:
@Lukasa How about add a specific lock to synchronize stream creation/deletion instead of using `_lock`? If it is so, we can acquire it when we already held `_write_lock` or `_read_lock` without deadlock issue.

Lukasa:
@plucury That's what `_lock` is: it's the specific lock that synchronizes stream creation/deletion. For a related reason it also synchronizes connection creation.

plucury:
fixed

PR #280, opened 2016-09-19
try to avoid deadlock

plucury:
Ref #275 

For avoid possible deadlock, I just add a `_OrderedRLock` object which will hold its inner lock when you try to acquire it. That's to say if you hold `_write_lock` you hold `_read_lock` too or if you hold `_lock` you hold `_write_lock` and `_read_lock` too. 

I'm not sure if this is overdoing it a little bit. What do you think about this? @Lukasa

### DIFF
--- a/hyper/http20/connection.py
+++ b/hyper/http20/connection.py
@@ -139,36 +139,9 @@ class HTTP20Connection(object):
 
         # Concurrency
         #
-        # Use one lock (_lock) to synchronize any interaction with global
-        # connection state, e.g. stream creation/deletion.
-        #
-        # It's ok to use the same in lock all these cases as they occur at
-        # different/linked points in the connection's lifecycle.
-        #
-        # Use another 2 locks (_write_lock, _read_lock) to synchronize
-        # - _send_cb
-        # - _recv_cb
-        # respectively.
-        #
-        # I.e, send/recieve on the connection and its streams are serialized
-        # separately across the threads accessing the connection.  This is a
-        # simple way of providing thread-safety.
-        #
-        # _write_lock and _read_lock synchronize all interactions between
-        # streams and the connnection.  There is a third I/O callback,
-        # _close_stream, passed to a stream's constructor.  It does not need to
-        # be synchronized, it uses _send_cb internally (which is serialized);
-        # its other activity (safe deletion of the stream from self.streams)
-        # does not require synchronization.
-        #
-        # _read_lock may be acquired when already holding the _write_lock,
-        # when they both held it is always by acquiring _write_lock first.
-        #
-        # Either _read_lock or _write_lock may be acquired whilst holding _lock
-        # which should always be acquired before either of the other two.
+        # Use one universal lock (_lock) to synchronize all interaction
+        # with global connection state, _send_cb and _recv_cb.
         self._lock = threading.RLock()
-        self._write_lock = threading.RLock()
-        self._read_lock = threading.RLock()
 
         # Create the mutable state.
         self.__wm_class = window_manager or FlowControlManager
@@ -232,7 +205,7 @@ class HTTP20Connection(object):
         :returns: Nothing
         """
         self.connect()
-        with self._write_lock:
+        with self._lock:
             with self._conn as conn:
                 conn.ping(to_bytestring(opaque_data))
             self._send_outstanding_data()
@@ -271,7 +244,7 @@ class HTTP20Connection(object):
         # If threads interleave these operations, it could result in messages
         # being sent in the wrong order, which can lead to the out-of-order
         # messages with lower stream IDs being closed prematurely.
-        with self._write_lock:
+        with self._lock:
             stream_id = self.putrequest(method, url)
 
             default_headers = (':method', ':scheme', ':authority', ':path')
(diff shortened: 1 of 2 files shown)


test/test_integration.py
```diff
@@ -935,6 +936,43 @@ def socket_handler(listener):
...
+        pool = ThreadPoolExecutor(max_workers=2)
+        pool.submit(do_connect, conn)
+        f = pool.submit(do_req, conn)
+
+        try:
+            f.result(timeout=10)
+        except TimeoutError:
+            assert False
+
+        self.tear_down()
```
Lukasa, line 1128: I'd love to see a comment on this test that points out that it's fundamentally probabilistic: it'll only fail transiently, and that we need to accept that. That will help us take it more seriously if it _does_ transiently fail.

hyper/http20/connection.py
```diff
@@ -161,9 +182,9 @@ def __init__(self, host, port=None, secure=None, window_manager=None,
         #
         # Either _read_lock or _write_lock may be acquired whilst holding _lock
         # which should always be acquired before either of the other two.
-        self._lock = threading.RLock()
-        self._write_lock = threading.RLock()
         self._read_lock = threading.RLock()
+        self._write_lock = _OrderedRLock(self._read_lock)
+        self._lock = _OrderedRLock(self._write_lock)
```
Lukasa: I think this might be backwards. The rule is that `_read_lock` requires that you hold `_write_lock` and acquire `_write_lock` first, but this code gets it backwards. The same is true for `_lock`, which always needs to be held first.

This means that I think the correct declaration is:

``` python
self._lock = threading.RLock()
self._write_lock = _OrderedRLock(self._lock)
self._read_lock = _OrderedRLock(self._write_lock)
```

Lukasa: There's a bigger problem here, which we need to address. That's the fact that this necessarily means that now to take the `_read_lock` you have to take _all_ the locks: that means that our _outer_ lock, which normally exists to act as the minimum level of contention, is now impossible to hold when someone is holding the write lock. That rather defeats the point of having this level of lock granularity.

I think we need another approach here. =(

plucury:
@Lukasa Yes, granularity, that's what I concern too. I declared locks just like what you suggested above when I start this PR, then I thought maybe it can be backward.

| lock aquired | forward | backward |
|-------------|---------|-----------|
| read_lock  | hold all three lock| hold read_lock only  |
| write_lock | hold both write_lock and lock  | hold write_lock and read_lock |
| lock | hold lock only | hold all three lock|

According to above table, If we only consider about the final state of acquiring operation, you will find the backward declaration is more approach the original idea of these locks.

However, I think the key point is that we can't know if it needs acquire its outer lock when we try to acquiring a lock. For example, when a process acquiring the read_lock to read socket, it doesn't know if it is necessary  to acquiring the write_lock at that time since it depends on what it read. Thus we can't always avoid acquiring read_lock before write_lock and maybe it's a good idea to make sure a process can acquire read_lock before it acquiring write_lock?

Lukasa:
Annoyingly, there are sort of four locks here which need synchronising. Frankly, at a certain point we start getting altogether quite tricky. @plucury are you proposing that the lock inversion shown in this patch is sufficient?

plucury:
@Lukasa  Yes, I think we can't ensure these locks always be acquired in order. But we can avoid deadlock even they were acquired in wrong order.

Lukasa:
Hrm, I feel like the lock inversion here is an awkward middle ground between having fine-grained locks and just choosing to have a single lock that protects the entire connection object. If we're going to have the complexity of managing a three-level lock hierarchy, I think we should do our best to avoid needing to acquire all three locks any time we want to read. On the other hand, if we *do* require that then we should consider whether we just want each connection to be locked behind a single lock, which has the advantage of giving us much stronger guarantees on behaviour.

Lukasa:
Oh, and @tbetbetbe may well have opinions here, I'll tag him in!

plucury:
@Lukasa I just have another try. Instead of using specific lock, I'm trying to keep acquiring locks in order this time. What do you think about it?

hyper/http20/connection.py
```diff
@@ -266,8 +266,9 @@ def request(self, method, url, body=None, headers=None):
         # If threads interleave these operations, it could result in messages
         # being sent in the wrong order, which can lead to the out-of-order
         # messages with lower stream IDs being closed prematurely.
+        stream_id = self.putrequest(method, url)
+        self.connect()
```
Lukasa: Sadly, this won't work either, as noted in the comment above.

`putrequest` obtains a stream ID, but doesn't send a HEADERS frame. That means that hyper thinks the stream ID is used, but the remote peer doesn't think that yet.

This whole problem is sufficiently troublesome that I suspect it reflects a more profound problem with the way this library is structured. It is extremely not amenable to being thread-safe except at an extremely coarse level.

I wonder if we need to investigate rewriting this library with threads. =(

plucury: @Lukasa Ah, I see. That's true. So how about simply use the single lock model at the moment?

Lukasa: I think the single lock model is likely to be the most usable approach at this time.

plucury:
@Lukasa I just transfer connection to single lock model. But I'm not sure why it breaks tests. Would you please have a look? Thanks a lot!

nateprewitt:
@plucury this is a bug in hyper-h2==2.5.0 (#291) which has been patched (python-hyper/hyper-h2#382) but isn't in a released version. I don't believe that it's related to your changes.

plucury:
Ah, I see. Thanks for your information @nateprewitt .

vinnyspb:
Hi guys,

Any reason not to merge this pull request with single lock?
I can confirm that deadlock issue is absolutely real and we've hit this problem in production.

Also I can confirm that the single lock fixes the problem.

Please let me know if I could help somehow.

Lukasa:
@vinnyspb Right now the reason is "because it breaks the tests" =)

plucury:
@Lukasa Just make it up to date. Please review again.

Issue #291, opened 2016-10-24
Hyper not stripping Connection header field


persiaAziz-zz:
Hyper sends TCP-FIN to the latest build of apache traffic server right after sending the request data if I mention 'Connection':'Keep-Alive' in my request header.

Lukasa:
@persiaAziz I'm afraid I'll need a little bit more information. Can you provide the code you're using please?

persiaAziz-zz:
``` python

import json
from hyper import HTTPConnection
import hyper
import argparse

def getResponseString(response):
    string = "HTTP/2 {0}\r\n".format(response.status)
    string+='date: '+response.headers.get('date')[0].decode('utf-8')+"\r\n"
    string+='server: '+response.headers.get('Server')[0].decode('utf-8')+"\r\n"
    return string

def makerequest(port):
    hyper.tls._context = hyper.tls.init_context()
    hyper.tls._context.check_hostname = False
    hyper.tls._context.verify_mode = hyper.compat.ssl.CERT_NONE

    conn = HTTPConnection('localhost:443'.format(port), secure=True)

    sites={'www.example.com'}
    #print("port---------",port)
    responses = []
    request_ids = []
    hdr={'Host':'www.example.com','Connection':'Keep-Alive'}
    for site in sites:
            request_id = conn.request('GET',url=site,headers=hdr)
            request_ids.append(request_id)

        # get responses
    for req_id in request_ids:
        response = conn.get_response(req_id)    
        print(getResponseString(response))


def main():
    parser = argparse.ArgumentParser()
    parser.add_argument("--port","-p",
                        type=int,                        
                        help="Port to use")
    args=parser.parse_args()
    makerequest(args.port)

if __name__ == '__main__':
    main()

```

persiaAziz-zz:
You can reproduce the issue with the above code. It is the Connection field in the header that causes the issue.

Lukasa:
@persiaAziz Do you get an exception in this code? I suspect you will.

persiaAziz-zz:
Yes I do @Lukasa it tells me 'stream forcefully closed' as I try to read the response

Lukasa:
Can you print the full stdout?

persiaAziz-zz:
`python3.5 h2client.py -p 443

Traceback (most recent call last):
  File "h2client.py", line 46, in <module>
    main()
  File "h2client.py", line 43, in main
    makerequest(args.port)
  File "h2client.py", line 31, in makerequest
[...]

Issue #285, opened 2016-09-28
Add a flag to connection and to response objects to find out connection type.


fendouai:
I can not get the http version from headers,how to get the http version when i use http2.

Lukasa:
@fendouai If you call `type()` on the response object you're holding, that will give you the object, which will be either a `HTTP20Response` or a `HTTP11Response`. That is enough to know the version.

However, I think we should add a feature that lets you consult a flag.

fendouai:
Thank you.  And I think the flag which shows the version of the http is important.

nateprewitt:
Would you like to help contribute a patch for this @fendouai? If not, I'd be willing to pick this up.

fendouai:
I am trying to find the version. I am not familiar with the project yet.If I have a solution,I will try to contribute a patch for this.

nateprewitt:
@Lukasa I've had this sitting for a while (9286249). If we still want it, I can open the PR. I wasn't sure if standard practice was truncating ".0" from HTTP/2 or if we even wanted "HTTP" in the version flag.

Lukasa:
I actually strongly recommend using the `enum` module and then using an `enum` to flag this. Let's pretend we're real programmers. ;)

Lukasa:
Resolved by #296.

PR #296, opened 2016-12-01
add HTTP version to connection and responses

nateprewitt:
This should address #285.

The tests aren't passing right now because we still have an outstanding bug (#291) in hyper until a new version of h2 is released. Other than the b'connection' header failures though, everything seems to be working.

### DIFF
--- a/hyper/http20/connection.py
+++ b/hyper/http20/connection.py
@@ -14,7 +14,9 @@ from ..tls import wrap_socket, H2_NPN_PROTOCOLS, H2C_PROTOCOL
 from ..common.exceptions import ConnectionResetError
 from ..common.bufsocket import BufferedSocket
 from ..common.headers import HTTPHeaderMap
-from ..common.util import to_host_port_tuple, to_native_string, to_bytestring
+from ..common.util import (
+    to_host_port_tuple, to_native_string, to_bytestring, HTTPVersion
+)
 from ..compat import unicode, bytes
 from .stream import Stream
 from .response import HTTP20Response, HTTP20Push
@@ -91,6 +93,9 @@ class HTTP20Connection(object):
         and one also isn't provided in the ``proxy`` parameter, defaults to
         8080.
     """
+
+    version = HTTPVersion.http20
+
     def __init__(self, host, port=None, secure=None, window_manager=None,
                  enable_push=False, ssl_context=None, proxy_host=None,
                  proxy_port=None, force_proto=None, **kwargs):
--- a/test/test_hyper.py
+++ b/test/test_hyper.py
@@ -16,7 +16,7 @@ from hyper.http20.util import (
     combine_repeated_headers, split_repeated_headers, h2_safe_headers
 )
 from hyper.common.headers import HTTPHeaderMap
-from hyper.common.util import to_bytestring
+from hyper.common.util import to_bytestring, HTTPVersion
 from hyper.compat import zlib_compressobj, is_py2
 from hyper.contrib import HTTP20Adapter
 import hyper.http20.errors as errors
@@ -81,6 +81,10 @@ class TestHyperConnection(object):
         assert c.port == 443
         assert c.proxy_host == 'ffff:aaaa::1'
         assert c.proxy_port == 8443
+
+    def test_connection_version(self):
+        c = HTTP20Connection('www.google.com')
+        assert c.version is HTTPVersion.http20
 
     def test_ping(self, frame_buffer):
         def data_callback(chunk, **kwargs):
@@ -1096,6 +1100,10 @@ class TestResponse(object):
             received += chunk
 
         assert received == b'this is test data'
+
+    def test_response_version(self):
+        r = HTTP20Response(HTTPHeaderMap([(':status', '200')]), None)
+        assert r.version is HTTPVersion.http20
 
 
 class TestHTTP20Adapter(object):
--- a/test/test_integration.py
+++ b/test/test_integration.py
@@ -15,6 +15,7 @@ import pytest
 from h2.frame_buffer import FrameBuffer
 from hyper.compat import ssl
 from hyper.contrib import HTTP20Adapter
+from hyper.common.util import HTTPVersion
 from hyperframe.frame import (
     Frame, SettingsFrame, WindowUpdateFrame, DataFrame, HeadersFrame,
     GoAwayFrame, RstStreamFrame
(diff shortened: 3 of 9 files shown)


nateprewitt:
Actually it looks like TestImportPython2.test_cannot_import_python_2 is failing too because it raises an AttributeError rather than an ImportError.

This seems unrelated to this patch though. I can dig a bit if you don't immediately know what's causing that @Lukasa.

hyper/http20/connection.py
```diff
@@ -102,6 +104,9 @@ def __init__(self, host, port=None, secure=None, window_manager=None,
         else:
             self.host, self.port = host, port
 
+        # HTTP Version of the connection
+        self.version = HTTPVersion.http20
```
Lukasa: Given that these are static and don't change, let's set these on the class object rather than do this at initialisation time.

test/test_hyper.py
```diff
@@ -1097,6 +1101,10 @@ def test_read_compressed_frames(self):
 
         assert received == b'this is test data'
 
+    def test_response_version(self):
+        r = HTTP20Response(HTTPHeaderMap([(':status', '200')]), None)
+        assert r.version is HTTPVersion.http20
+
```
Lukasa, line 1107: Can we also get a few tests for the abstract HTTPConnection object passing this through?

Lukasa (changes requested):
This generally seems fine! I'll try to chase the test stuff up tomorrow.

nateprewitt:
A couple quick questions on the abstract tests. I'm currently unable to come up with a way to test HTTP and TLS upgrades without live network IO to something like http2bin.org. I could mock the connections like the other tests but that kind of defeats the point.

The second question I have is more of an oddity. I've written what is a successful test when run with py.test and as a separate python file, but in tox it repeatedly fails to perform the TLS upgrade. `wrap_socket` never changes the proto from `None`.

```python
from hyper.common.util import HTTPVersion

c = HTTPConnection('www.http2bin.org', 443)
assert c.version is HTTPVersion.http11
c.request('GET', '/')
assert c.version is HTTPVersion.http20
```

test/test_abstraction.py
```diff
@@ -73,6 +74,13 @@ def test_http_upgrade(self, monkeypatch):
         assert isinstance(c._conn, DummyH2Connection)
         assert c._conn._sock == 'totally a non-secure socket'
 
+    def test_HTTPConnection_version(self):
+        c = HTTPConnection('www.http2bin.org', 80)
+        assert c.version is HTTPVersion.http11
+        c.request('GET', '/')
+        c.get_response()
```
nateprewitt: Note: test the response version here too.

Lukasa:
> I'm currently unable to come up with a way to test HTTP and TLS upgrades without live network IO to something like http2bin.org. I could mock the connections like the other tests but that kind of defeats the point.

I don't think I follow this. Other parts of the test suite have found a way to do it, it should be possible for your code to do that. In this case, you just need to lie about the connection setup, which ought not to be too difficult.

>  I've written what is a successful test when run with py.test and as a separate python file, but in tox it repeatedly fails to perform the TLS upgrade. wrap_socket never changes the proto from None.

Are you running the same Python version in both cases? Same dependencies? Almost certainly you're using a Python that has a too-old ssl library or OpenSSL in tox, so there is no negotiation.

Lukasa:
Ok, the master branch should have its tests fixed up.

nateprewitt:
> I don't think I follow this. Other parts of the test suite have found a way to do it, it should be possible for your code to do that. In this case, you just need to lie about the connection setup, which ought not to be too difficult.

Sorry, I should have been clearer. The other parts of the library are mocking out a custom Connection object in order to do this. In that case, I'm simply creating a DummySocket that has an attr "version" and then testing the dummy socket still has a version after upgrades. That kind of test never actually touches the underlying `HTTP11Connection`  or `HTTP20Connection` so it doesn't prove anything useful. It will always pass regardless of changes to the actual connection classes.

>Are you running the same Python version in both cases? Same dependencies? Almost certainly you're using a Python that has a too-old ssl library or OpenSSL in tox, so there is no negotiation.

That was the first thing I had checked but couldn't find anything immediately. The weird thing is it will run fine locally in a fresh virtualenv, but fails in tox and Travis. I'll try to do a deeper dig to figure out what my base environment has that doesn't seem to be in Travis.

Lukasa:
Hrm. Can you do it in the socket-level style by actually creating a TLS connection? If not, we may need to use some mocks to achieve what we need.

nateprewitt:
Yeah, I'll do some tinkering this afternoon and try to devise a way to initiate the upgrades with the actual objects but no external IO.

test/test_abstraction.py
```diff
@@ -74,6 +77,85 @@ def test_http_upgrade(self, monkeypatch):
...
+
+            send_event.wait()
+
+            # We need to send back a response.
+            resp = (
+                b'HTTP/1.1 101 Switching Protocols\r\n'
+                b'Server: socket-level-server\r\n'
+                b'Content-Length: 0\r\n'
+                b'Connection: upgrade\r\n',
+                b'Upgrade: TLS/1.0, HTTP/2.0'
```
Lukasa: Uh...what is this? This test doesn't make any sense at all, and frankly suggests that the upgrade code in hyper is broken. =P

Lukasa (changes requested):
So this looks good, except for the fact that one of the tests is super-wacky and I don't really understand what it was trying to do.

test/test_abstraction.py
nateprewitt: Sorry, this one was a test based off of a bad reading of the RFC 2817. I had a question on it yesterday that I was going to ask here but solved it and forgot to remove the code.

nateprewitt:
@Lukasa, okay, sorry that took so long. I got the test working by monkeypatching `wrap_socket`. Things should be ready for an actual look now.

nateprewitt:
As an aside, is there a reason the tests are configured to pass despite failures, as long as there's 100% test coverage?

I hit this with a couple test attempts that were failing but the build wasn't explicitly telling me until I dug into the logs of "successful" builds. It looks like all of the NGHTTP2 builds have been failing to import nghttp2 (`TestUtilities.test_nghttp2_installs_correctly`) for at least the last few months too.

Lukasa:
There is not, I think it's just a misconfiguration of Travis. I seem to recall that I had the same problem in the hpack library, which I fixed in python-hyper/hpack#60. This included removing the nghttp2 support (which hasn't worked for a while, and which I didn't have the resources to investigate).

I recommend opening a PR to do both of those things if you have the bandwidth to do it. =)

test/test_integration.py
```diff
@@ -935,6 +936,100 @@ def socket_handler(listener):
...
+            # Send the headers for the response. This response has no body.
+            f = build_headers_frame(
+                [(':status', '200'), ('content-length', '0')]
+            )
+            f.flags.add('END_STREAM')
+            f.stream_id = 1
+            sock.sendall(f.serialize())
+
+            # Wait for the message from the main thread.
+            send_event.set()
```
Lukasa: This needs to be `wait`, not `set`.

Lukasa (changes requested):
Small notes, but I think it'd be good if we tackled the bad reporting of tests first. =)

Lukasa:
Nevermind, I'm tackling it myself in #298.

nateprewitt:
Alright, things should be set once the tests finish.

Lukasa (approved):
Looks good to me, thanks @nateprewitt!

PR #301, opened 2017-01-16
Support generator type data

xiongyihui:
1. support generator type data

   Usage:
   ```
   from hyper import HTTPConnection

   def gen():
        yield 'hello'
        yield 'world'

   conn = HTTPConnection('http2bin.org:443')
   conn.request('POST', '/post', body=gen())
   print(conn.get_response().read())
   ```

2. fix that data with integral multiple of the `MAX_CHUNK` size does not end stream.

### DIFF
--- a/hyper/http20/stream.py
+++ b/hyper/http20/stream.py
@@ -13,6 +13,7 @@ stream is an independent, bi-directional sequence of HTTP headers and data.
 Each stream is identified by a monotonically increasing integer, assigned to
 the stream by the endpoint that initiated the stream.
 """
+import types
 import h2.exceptions
 
 from ..common.headers import HTTPHeaderMap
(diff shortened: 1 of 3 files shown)


hyper/http20/stream.py
```diff
@@ -107,23 +108,26 @@ def send_data(self, data, final):
...
+
+            # If ``final`` is True, send a empty chunk to end stream
+            if final:
+                self._send_chunk('', final)
+        elif hasattr(data, 'read'):
             while True:
-                data = fobj.read(MAX_CHUNK)
-                yield data
+                chunk = fobj.read(MAX_CHUNK)
                 if len(data) < MAX_CHUNK:
```
Lukasa: You've changed the variable name here, so this is almost certainly not going to work.

Lukasa (changes requested):
Thanks for this change! While it's valuable, I'm afraid that I need tests for all the changes you've made before I can merge it: that is, we need to test the new generator upload function, and we also need tests for the CHUNK_LENGTH bug that fail before this patch and work after. =)

xiongyihui:
My bad. Updated. It still failed :(

PR #302, opened 2017-01-17
Fix integration test for Python 3

laike9m:
Originally this two tests `test_adapter_received_values` and `test_adapter_sending_values` would fail in Python 3 due to `connection` header not being removed, which is a bug of the `h2_safe_headers` function. 

`headers(class HTTPHeaderMap)` stores bytes because we call `to_bytestring_tuple` when initiating it. While in Python 2 `bytes` is equivalent to `str`, and `h2_safe_headers` can find `'connection'` successfully, in Python 3 `b'connection' != 'connection'`, so this header is not removed.

Some affected tests are also modified.

### DIFF
--- a/hyper/http20/util.py
+++ b/hyper/http20/util.py
@@ -52,9 +52,9 @@ def h2_safe_headers(headers):
     """
     stripped = {
         i.lower().strip()
-        for k, v in headers if k == 'connection'
-        for i in v.split(',')
+        for k, v in headers if k == b'connection'
+        for i in v.split(b',')
     }
-    stripped.add('connection')
+    stripped.add(b'connection')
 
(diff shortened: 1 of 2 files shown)


Lukasa:
This looks good to me, thanks so much! :sparkles:

laike9m:
BTW how to join the Hyper organization, created certain amount of PRs?

Lukasa:
@laike9m *Strictly*, this repository is not a part of the Hyper organisation yet. I appreciate that that's confusing, but for various boring organisational reasons I haven't brought it in yet. However, literally any PR against any of the projects in [this org](https://github.com/python-hyper) would qualify you.

Issue #300, opened 2017-01-12
The _remaining_capacity > _bytes_in_buffer check in bufsocket.py is unnecessary


laike9m:
This is not an actual issue, but a possible tiny improvement.

**TL;DR**
I think this check `self._remaining_capacity > self._bytes_in_buffer` in [bufsocket.py#L141](https://github.com/Lukasa/hyper/blob/development/hyper/common/bufsocket.py#L141) is unnecessary.

Here's why.
From the definitions we know:
```python
  self._remaing_capacity - self._bytes_in_buffer
= (self._buffer_size - self._index) - self._bytes_in_buffer
= self._buffer_size - (self._index + self._bytes_in_buffer)
= self._buffer_size - self._buffer_end    
```

We also know that each time `self._bytes_in_buffer` increases, the increment is less than 
  `_buffer_size - _buffer_end`. An example:
```python
if (self._remaining_capacity > self._bytes_in_buffer and should_read):
    count = self._sck.recv_into(self._buffer_view[self._buffer_end:])
    if not count and amt > self._bytes_in_buffer:
        raise ConnectionResetError()
    self._bytes_in_buffer += count   # count cannot be greater than _buffer_size - _buffer_end
```

Now we have two formulas:
```
_buffer_size - _buffer_end = _remaing_capacity - _bytes_in_buffer    ①
count < _buffer_size - _buffer_end                                   ②   
```

Thus `_bytes_in_buffer < _remaing_capacity` is always `True`.
```
  _bytes_in_buffer(new) 
= _bytes_in_buffer(old) + count 
< _bytes_in_buffer + _buffer_size - _buffer_end 
= _remaing_capacity
```

I intended to remove the check and run tests, but unfortunately I was not able to finish them(I have all the test deps installed and run `make test`, don't know why it takes so long). So I'm not 100% sure it can be removed.

Lukasa:
I suspect you're right. ;) Anyone who wants to apply this patch is welcome to.

laike9m:
I wonder how long do all the tests take to run, normally?

Lukasa:
For running the tests I recommend using tox, but you may have encountered a test hang. This project has CI, so you can simply open a PR if you'd prefer.

laike9m:
Could you give me an estimate? So that I know if tests indeed hang. `pytest-xdist` shows I'm using all 4 cores, I don't know whether using tox can speed it up.

Lukasa:
[...]

PR #303, opened 2017-01-18
Removed unnecessary size check

laike9m:
#300


Lukasa:
Looks safe to me, thanks @laike9m! :sparkles:

PR #305, opened 2017-01-24
More informative error message if no protocol found for http2

ojii:
Today I ran into an issue with hyper, where the OpenSSL version in my environment was too old (1.0.1) so it did not have ALPN, which was required by the server. This is of course not a bug/problem with hyper itself, however the error message I got from hyper was not that helpful (it was an `AssertionError` with no help message). This pull request suggests a new check after wrapping the socket in SSL to see if a protocol was selected and if not tells the user that none was selected and suggests the user check their OpenSSL version.

### DIFF
--- a/test/test_http20.py
+++ b/test/test_http20.py
@@ -1,0 +1,42 @@
+# -*- coding: utf-8 -*-
+"""
+test_http20.py
+~~~~~~~~~~~~~~
+
+Unit tests for hyper's HTTP/2.0 implementation.
+"""
+import pytest
+from mock import patch
+
+from server import SocketLevelTest
+
+
+class TestHTTP20Connection(SocketLevelTest):
+    h2 = True
+
+    def test_useful_error_with_no_protocol(self):
+        self.set_up()
+
+        def socket_handler(listener):
+            sock = listener.accept()[0]
+            sock.close()
+
+        self._start_server(socket_handler)
+        conn = self.get_connection()
+
+        with patch('hyper.http20.connection.wrap_socket') as mock:
+            mock.return_value = (None, None)
+            with pytest.raises(AssertionError) as exc_info:
+                conn.connect()
+        assert (
+            "No suitable protocol found."
+            in
+            str(exc_info)
+        )
+        assert (
+            "Check your OpenSSL version."
+            in
+            str(exc_info)
+        )
+
+        self.tear_down()
--- a/test/test_integration.py
+++ b/test/test_integration.py
@@ -12,6 +12,7 @@ import time
 import hyper
 import hyper.http11.connection
 import pytest
+from mock import patch
 from h2.frame_buffer import FrameBuffer
 from hyper.compat import ssl
 from hyper.contrib import HTTP20Adapter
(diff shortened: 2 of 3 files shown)


Lukasa:
@ojii Thanks for this patch, this looks like a good change. Do you mind adding a test to confirm that the message is present? That way we won't accidentally regress it.

ojii:
@Lukasa will try. Had issues running the test suite (related to my network setup, not the test suite itself from what I can see), which is why there's no test included for now.

ojii:
@Lukasa I've added a test, though I had to mock out all of `wrap_socket`. I hope this is okay.

test/test_http20.py
```diff
@@ -0,0 +1,37 @@
...
+
+        with patch('hyper.http20.connection.wrap_socket') as mock:
+            mock.return_value = (None, None)
+            with pytest.raises(AssertionError) as exc_info:
+                conn.connect()
+                assert (
+                    "No suitable protocol found, check your OpenSSL version"
+                    in
+                    str(exc_info)
+                )
```
Lukasa: You can pull this assertion out of the context manager, which you will in fact need to do to get it to run because the `AssertionError` that gets thrown terminates the execution of this block.

test/test_http20.py
```diff
@@ -0,0 +1,37 @@
...
+
+        self.tear_down()
```
Lukasa: Newline at the end of the file please! =)

Lukasa (changes requested):
This is generally fine, just got a pair of notes here.

ojii:
Looks like https://github.com/ojii/hyper/blob/bdcf38e403c1f669d8d66d760ad16ee4c7153451/test/test_integration.py#L38breaks the other tests. `proto` becomes `None` in `wrap_socket`, but the next assertion that `proto` is in `H2_NPN_PROTOCOLS` passes because `None` is allowed.

Of course with my extra check, this fails now with the explicit check against `None`.

Not sure what the best way to proceed is.

Lukasa:
So, is there any reason not to extend the assertion below rather than add a new one? If not, you should assert that the protocol is in `H2_NPN_PROTOCOLS`.

test/test_integration.py
```diff
@@ -34,8 +35,10 @@
     hyper.tls._context.check_hostname = False
     hyper.tls._context.verify_mode = ssl.CERT_NONE
 
-    # Cover our bases because NPN doesn't yet work on all our test platforms.
-    hyper.http20.connection.H2_NPN_PROTOCOLS += ['', None]
+PROTOCOLS = hyper.http20.connection.H2_NPN_PROTOCOLS + ['', None]
+
+
+# Cover our bases because NPN doesn't yet work on all our test platforms.
```
Lukasa, line 38: Why did this get moved?

Lukasa (changes requested):
Looks like a comment got moved.

ojii:
> Looks like a comment got moved.

Really sorry about how messy this pull request got. Should I re-open another PR with the changes as one commit?

Lukasa:
There's no need: if you rebase the branch to squash the commits down and then force push to the same branch, GitHub will work it out. =D

ojii:
I hope it's fine now. Let me know if there's any other changes that are needed.

Lukasa (approved):

Lukasa:
Looks good to me. Good work @ojii! :sparkles:

ojii:
Thank you for your great work bringing HTTP2 to Python.

PR #307, opened 2017-02-11
Implement HTTP20Adapter.close to close connections

### LINKED ISSUE #306: HTTP20Adapter object has no attribute 'poolmanager'
Using `HTTP20Adapter` with `requests` and closing the session (with context manager) results in attribute error.

Usage:
```python
with requests.Session() as session:
        session.mount(current_app.config['APNS_SERVER'], hyper.contrib.HTTP20Adapter())
        response = session.post(...)
        response.raise_for_status()
```

Error:
```
    118                 print(response.request.headers)
    119             else:
--> 120                 print(response.headers[b'apns-id'])
    121 
    122         # try:

blaze/.venv/lib/python3.5/site-packages/requests/sessions.py in __exit__(self, *args)
    367 
    368     def __exit__(self, *args):
--> 369         self.close()
    370 
    371     def prepare_request(self, request):

blaze/.venv/lib/python3.5/site-packages/requests/sessions.py in close(self)
    688         """Closes all adapters and as such the session"""
    689         for v in self.adapters.values():
--> 690             v.close()
    691 
    692     def mount(self, prefix, adapter):

blaze/.venv/lib/python3.5/site-packages/requests/adapters.py in close(self)
    304         which closes any pooled connections.
    305         """
--> 306         self.poolmanager.clear()
    307         for proxy in self.proxy_manager.values():
    308             proxy.clear()

AttributeError: 'HTTP20Adapter' object has no attribute 'poolmanager'
```

Packages:
```
h2 (2.6.0.dev0)
hpack (2.3.0)
hyper (0.7.0)
hyperframe (3.2.0)
requests (2.12.1)
```

allanlei:
Fixes #306

### DIFF
--- a/test/test_hyper.py
+++ b/test/test_hyper.py
@@ -26,6 +26,7 @@ import pytest
 import socket
 import zlib
 from io import BytesIO
+import requests
 
 TEST_DIR = os.path.abspath(os.path.dirname(__file__))
 TEST_CERTS_DIR = os.path.join(TEST_DIR, 'certs')
@@ -1127,6 +1128,22 @@ class TestHTTP20Adapter(object):
             'http',
             cert=CLIENT_PEM_FILE)
         assert conn1 is conn2
+
+    def test_adapter_close(self):
+        """
+        Tests HTTP20Adapter properly closes connections
+        """
+        s = requests.Session()
+        s.mount('https://', HTTP20Adapter())
+        s.close()
+
+    def test_adapter_close_context_manager(self):
+        """
+        Tests HTTP20Adapter properly closes connections via context manager
+        """
+        with requests.Session() as s:
+            a = HTTP20Adapter()
+            s.mount('https://', a)
 
 
 class TestUtilities(object):
--- a/hyper/contrib.py
+++ b/hyper/contrib.py
@@ -157,3 +157,7 @@ class HTTP20Adapter(HTTPAdapter):
         orig.msg = FakeOriginalResponse(resp.headers.iter_raw())
 
         return response
+
+    def close(self):
+        for connection in self.connections.values():
+            connection._conn.close()


allanlei:
The `http2bin.org`  endpoint seems to be down, is there another HTTP2 endpoint or a way to dummy the connection?

test/test_hyper.py
```diff
@@ -1128,6 +1129,25 @@ def test_adapter_accept_client_certificate(self):
...
+        s.close()
+
+    def test_adapter_close_context_manager(self):
+        """
+        Tests HTTP20Adapter properly closes connections via context manager
+        """
+        with requests.Session() as s:
+            a = HTTP20Adapter()
+            s.mount('https://http2bin.org', a)
+            s.get('https://http2bin.org/')
```
Lukasa: I strongly recommend avoiding making real web requests in this test module (even if the rest of them do).

I recommend mocking out some of the lower-level functions. It's probably enough to mock out the `send` method of the adapter and the `HTTPConnection` object.

Lukasa (changes requested):
Cool, so this looks reasonable! I have made some suggestions for alternative testing approaches.

allanlei:
I've changed the test to just mount the adapter without making a request.  It raises the same error.

test/test_hyper.py
allanlei: I've changed the test to just mount the adapter without making a request. It raises the same error.

nateprewitt: Hey @allanlei, it looks like pytest doesn't think this test case is actually hitting line 163 (`connection._conn.close()`) in contrib.py. Until that's properly triggered, Travis will continue returning a failure.

Without initiating a request, I'm not sure if `self.connections.values()` will be populated. If that's the case, the for loop will never execute.

Lukasa:
So I think @nateprewitt's deleted comment is right: I think we need to initiate a HTTP/2 request here. Rather than use http2bin you may need to write a socket-level test as is done elsewhere in the code.

nateprewitt:
Not deleted, just hiding on the last "requested changes" comment thread 😉

PR #309, opened 2017-02-21

hkwi:
Push flag required for the case the initial upgrade request triggered server push.


hkwi:
Closing because I found a bug that does not pass enable_push kwargs

PR #310, opened 2017-02-21

hkwi:
Push flag required for the case the initial upgrade request triggered server push.

### DIFF
--- a/hyper/http11/connection.py
+++ b/hyper/http11/connection.py
@@ -278,8 +278,9 @@ class HTTP11Connection(object):
         http2_settings = SettingsFrame(0)
         http2_settings.settings[SettingsFrame.INITIAL_WINDOW_SIZE] = 65535
         if self._enable_push is not None:
-            http2_settings.settings[
-                SettingsFrame.ENABLE_PUSH] = int(self._enable_push)
+            http2_settings.settings[SettingsFrame.ENABLE_PUSH] = (
+                int(self._enable_push)
+            )
         encoded_settings = base64.urlsafe_b64encode(
             http2_settings.serialize_body()
         )
@@ -352,7 +353,7 @@ class HTTP11Connection(object):
         Handles streaming a file-like object to the network.
         """
         while True:
-            block = fobj.read(16*1024)
+            block = fobj.read(16 * 1024)
             if not block:
                 break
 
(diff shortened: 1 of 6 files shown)


Lukasa (changes requested):
Thanks for this!

This patch is reasonable, but there are a bunch of tests failing that need to be updated to take account of this change. Do you mind doing that?

hkwi:
I'll look into.

hyper/http11/connection.py
```diff
@@ -276,6 +277,9 @@ def _add_upgrade_headers(self, headers):
         # Settings header.
         http2_settings = SettingsFrame(0)
         http2_settings.settings[SettingsFrame.INITIAL_WINDOW_SIZE] = 65535
+        if self._enable_push is not None:
+            http2_settings.settings[
+                SettingsFrame.ENABLE_PUSH] = int(self._enable_push)
```
Lukasa: Rather than break here, can you do this?

```python
http2_settings.settings[SettingsFrame.ENABLE_PUSH] = (
    int(self._enable_push)
)
```

Lukasa (changes requested):
So you've fixed some of the test failures, but you're now not actually testing that this data gets emitted. We do need to add a new test to cover this branch.

hyper/http11/connection.py
hkwi: Yes.

Lukasa (changes requested):
So, as noted, this is not really tested in any sense right now. It would be good to see the appropriate testing changes made.

hkwi:
One travis-ci job is failing unexpectedly with empty raw log. Can I ask help?

Lukasa:
I've restarted that build. =)

hkwi:
I added http11 h2c upgrading test by inheriting TestServerPush which is for http20 h2, and test passed.

Lukasa (changes requested):
So, I'm not sure about this approach.

You've added `getpushes` to the `HTTP11Connection` object with the goal of having it be a dummy method for catching the upgrade. That's understandable, but I'm not sure that this is the way users should be interacting with the abstract method: in particular because if the connection *does not* upgrade, which the user cannot guarantee will not happen, the response object is lost.

For the purpose of testing it's probably enough to call `getresponse` and then check for pushes. How does that sound?

hkwi:
I agree that `HTTP11Connection.getpushes` is a little bit hacky. This was added to meet the strict tests of `test_promise_before_headers` or `test_capture_all_promises`. If we don't need to satisfy these tests for http11 upgrade, implementation would be more relaxed style. I'll create another PR for that version later.
BTW, the response will be stored in `self._response` for later `self.get_response` call, and would not lost.

Lukasa:
I think those tests don't need to work for upgrade, but we should write a test that *does* work as expected (e.g. you get a HTTP/2 response and *then* look for pushes on that stream)

PR #311, opened 2017-02-23

hkwi:
Push flag required for the case the initial upgrade request triggered server push. This PR is a relaxed version of #310 , which was more strict on server push I/O timing.
Test send/recv timing was fixed, because connection state machine does not get corrupted during h2c upgrade migration.

### DIFF
--- a/test/test_hyper.py
+++ b/test/test_hyper.py
@@ -48,7 +48,6 @@ def frame_buffer():
 
 
 class TestHyperConnection(object):
-
     def test_connections_accept_hosts_and_ports(self):
         c = HTTP20Connection(host='www.google.com', port=8080)
         assert c.host == 'www.google.com'
@@ -500,7 +499,6 @@ class TestHyperConnection(object):
 
     def test_send_tolerate_peer_gone(self):
         class ErrorSocket(DummySocket):
-
             def sendall(self, data):
                 raise socket.error(errno.EPIPE)
 
@@ -703,7 +701,6 @@ class TestHyperConnection(object):
 
 
 class FrameEncoderMixin(object):
-
     def setup_method(self, method):
         self.frames = []
         self.encoder = Encoder()
@@ -737,7 +734,6 @@ class FrameEncoderMixin(object):
 
 
 class TestServerPush(FrameEncoderMixin):
-
     def request(self, enable_push=True):
         self.conn = HTTP20Connection('www.google.com', enable_push=enable_push)
         self.conn._sock = DummySocket()
@@ -963,7 +959,6 @@ class TestServerPush(FrameEncoderMixin):
 
 
 class TestResponse(object):
-
     def test_status_is_stripped_from_headers(self):
         headers = HTTPHeaderMap([(':status', '200')])
         resp = HTTP20Response(headers, None)
@@ -1114,7 +1109,6 @@ class TestResponse(object):
 
 
 class TestHTTP20Adapter(object):
-
     def test_adapter_reuses_connections(self):
         a = HTTP20Adapter()
         conn1 = a.get_connection('http2bin.org', 80, 'http')
@@ -1138,7 +1132,6 @@ class TestHTTP20Adapter(object):
 
 
 class TestUtilities(object):
-
     def test_combining_repeated_headers(self):
         test_headers = [
             (b'key1', b'val1'),
@@ -1466,7 +1459,6 @@ class TestUpgradingPush(FrameEncoderMixin):
 
 # Some utility classes for the tests.
 class NullEncoder(object):
-
     @staticmethod
     def encode(headers):
 
@@ -1483,7 +1475,6 @@ class NullEncoder(object):
 
 
 class FixedDecoder(object):
-
     def __init__(self, result):
         self.result = result
 
@@ -1492,7 +1483,6 @@ class FixedDecoder(object):
 
 
 class DummySocket(object):
-
     def __init__(self):
         self.queue = []
         self._buffer = BytesIO()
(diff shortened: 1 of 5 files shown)


test/test_hyper.py
```diff
@@ -1303,6 +1303,188 @@ def test_resetting_streams_after_close(self):
...
+        if end_stream:
+            frame.flags.add('END_STREAM')
+        self.frames.append(frame)
+
+    def add_data_frame(self, stream_id, data, end_stream=False):
+        frame = DataFrame(stream_id)
+        frame.data = data
+        if end_stream:
+            frame.flags.add('END_STREAM')
+        self.frames.append(frame)
```
Lukasa: Can we move these helpers out to a common mixin class now that we're using them in more than one place?

test/test_hyper.py
```diff
@@ -1303,6 +1303,188 @@ def test_resetting_streams_after_close(self):
...
+
+    def request(self, enable_push=True):
+        self.conn = HTTPConnection('www.google.com', enable_push=enable_push)
+        self.conn._conn._sock = DummySocket()
+        self.conn._conn._sock.buffer = BytesIO(
+            self.http101 + b''.join([frame.serialize()
+                                     for frame in self.frames])
+        )
+        self.conn.request('GET', '/')
```
Lukasa, line 1322: I wonder if this should do any asserting that the Upgrade header is present?

test/test_integration.py
```diff
@@ -958,16 +1004,15 @@ def socket_handler(listener):
...
             # Send the headers for the response. This response has no body.
             f = build_headers_frame(
                 [(':status', '200'), ('content-length', '0')]
             )
             f.flags.add('END_STREAM')
             f.stream_id = 1
             sock.sendall(f.serialize())
-
-            # Wait for the message from the main thread.
-            send_event.wait()
```
Lukasa: Why did this get moved?

test/test_integration.py
```diff
@@ -1078,8 +1126,6 @@ def socket_handler(listener):
             d.data = b'1234567890' * 2
             d.flags.add('END_STREAM')
             sock.send(d.serialize())
-
-            send_event.wait(5)
```
Lukasa: Why was this removed?

Lukasa (changes requested):
So this seems reasonable, but there are some changes in the tests I don't understand. Notes are inline.

test/test_hyper.py
hkwi: I got it.

test/test_hyper.py
hkwi: Upgrading process itself is tested at `test_http11.py`, and IMHO, we would focus on client side view here.

Lukasa: Yeah, that's fair enough.

test/test_integration.py
hkwi: We have to wait here only when the client side are going to get the response. Another option is add `send_event.wait()` here, and add `get_response()`, `send_event.set()` in client side.

test/test_integration.py
hkwi: We have to wait here when we close the socket. Another option is add `send_event.wait()` and `sock.close()` here, and `send_event.set()` in client side.

test/test_integration.py
Lukasa: So historically the reason for these waits has been to ensure that we don't close sockets in a way that leads to unexpected socket behaviour on some platforms. If we don't *need* to remove them I'd be inclined to leave them in.

test/test_integration.py
Lukasa: I don't think this really answered what the reason was for removing this. Was it causing a problem?

test/test_integration.py
hkwi: OK

test/test_integration.py
hkwi: It does not causing problem. I removed just because I thought it was not necessary.

Lukasa: Yeah, for this stuff that's not really a great attitude. Specifically, these are intended to delay socket shutdown to the point where the socket is no longer in use, as some interactions can cause exceptions to be raised if EOF is received too early.

test/test_hyper.py
```diff
@@ -1107,6 +1114,7 @@ def test_response_version(self):
 
 
 class TestHTTP20Adapter(object):
+
```
Lukasa: Why did all these blank lines get added?

Lukasa (changes requested):

test/test_hyper.py
hkwi: Ah... that's autopep8. I'd revert those.

test/test_integration.py
```diff
@@ -64,17 +65,30 @@ def frame_buffer():
...
+    buffer.data = data
+
+
 def receive_preamble(sock):
     # Receive the HTTP/2 'preamble'.
-    first = sock.recv(65535)
+    client_preface = b'PRI * HTTP/2.0\r\n\r\nSM\r\n\r\n'
+    timeout = time.time() + 5
+    got = b''
+    while len(got) < len(client_preface) and time.time() < timeout:
```
Lukasa: What's the logic behind this timeout here? The socket doesn't actually have the timeout applied to it, so it could time out indefinitely.

Lukasa (changes requested):
Ok, cool, getting closer. One more question inline.

test/test_integration.py
hkwi: Oh, that's a bug. I used timeout for local testing.

Lukasa (approved):
Cool, this looks good to me. Thanks for doing all this work @hkwi! :sparkles:

Issue #312, opened 2017-03-04
Doesn’t work with h2c on nghttp2.org


vfaronov:
Not sure if the problem is in Hyper/h2 or in nghttp2, but `AssertionError` would suggest the former:

```
$ hyper GET http://nghttp2.org/
Traceback (most recent call last):
  File "/home/vasiliy/tmp/env4/bin/hyper", line 11, in <module>
    load_entry_point('hyper==0.8.0.dev0', 'console_scripts', 'hyper')()
  File "/home/vasiliy/tmp/env4/lib/python3.5/site-packages/hyper/cli.py", line 259, in main
    data = request(args)
  File "/home/vasiliy/tmp/env4/lib/python3.5/site-packages/hyper/cli.py", line 249, in request
    response = conn.get_response()
  File "/home/vasiliy/tmp/env4/lib/python3.5/site-packages/hyper/common/connection.py", line 130, in get_response
    return self._conn.get_response(*args, **kwargs)
  File "/home/vasiliy/tmp/env4/lib/python3.5/site-packages/hyper/http11/connection.py", line 225, in get_response
    self
  File "/home/vasiliy/tmp/env4/lib/python3.5/site-packages/hyper/http11/response.py", line 69, in __init__
    assert self._expect_close or self._length is not None or self._chunked
AssertionError
```

With tcpdump, I see the HTTP/1.1 upgrade dance, and I see nghttp2.org sending the HTML. I can attach the dump if you wish.

This is with Git master versions of Hyper and h2.

Lukasa:
Yup, that's a simple bug: we don't properly do a case-insensitive compare of the Connection header.

This would be a fairly easy bug to fix: are you interested in fixing it? The comparison is on line 216 of hyper/http11/connection.py

vfaronov:
I would gladly contribute a fix, but I think that line should be removed altogether. `Connection: upgrade` does not mean “connection is being upgraded”, it means “the `Upgrade` header must not be forwarded by an intermediary” (see [RFC 7230 § 6.7](https://tools.ietf.org/html/rfc7230#section-6.7) and [§ 6.1](https://tools.ietf.org/html/rfc7230#section-6.1)), and that has no bearing on Hyper. A combination of status code and `Upgrade: h2c` is enough to proceed.

Lukasa:
You missed the relevant section of RFC 7230 section 6.7:

>  When Upgrade is sent, the sender MUST also send a Connection header field that contains an "upgrade" connection option

[...]

PR #313, opened 2017-03-04
Fix upgrading to h2c

vfaronov:
Fixes https://github.com/Lukasa/hyper/issues/312

### DIFF
--- a/test_release.py
+++ b/test_release.py
@@ -15,7 +15,8 @@ import logging
 import random
 import requests
 import threading
-from hyper import HTTP20Connection, HTTP11Connection
+from hyper import HTTP20Connection, HTTP11Connection, HTTPConnection
+from hyper.common.util import HTTPVersion
 from hyper.contrib import HTTP20Adapter
 
 logging.basicConfig(level=logging.INFO)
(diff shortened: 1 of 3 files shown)


test_release.py
```diff
@@ -149,3 +150,19 @@ def test_hitting_httpbin_org_http11(self):
...
+        c = HTTPConnection('nghttp2.org:80')
+
+        # Make the request.
+        c.request('GET', '/')
+        response = c.get_response()
+
+        # Check that the response is OK and that we did upgrade to HTTP/2.
+        assert response.status == 200
+        assert response.read()
+        assert response.version == HTTPVersion.http20
```
Lukasa, line 168: So this is a great test, thanks. But the release tests are generally allowed to fail because Travis sucks at running tests with real networks. Can you write a test in `test_integration.py` or `test_integration_http11.py` as well to ensure that this works properly?

Lukasa (changes requested):
Cool, this looks really good. I have one small requirement with the testing, but everything that's currently in the patch is 💯.

test_release.py
vfaronov: I did [change `test_integration_http11.py`](https://github.com/Lukasa/hyper/pull/313/files#diff-9e6d1edfa0b69f29474dd4ad10311941) in such a way that it fails without my fix. Do you think something else is necessary?

Lukasa: Oh, no, sorry, I just wasn't paying enough attention there. My bad!

PR #315, opened 2017-03-04
Document how to override certificate verification

vfaronov:

PR #316, opened 2017-03-05
Fix the documentation on overriding the SSL context

vfaronov:
Sorry, I completely botched #315. It seemed to work, but it was falling back to HTTP/1.1, which is definitely not what most people would want. And I missed `hyper.tls.init_context`. Hopefully this version is more correct, but it’s not like I know what I’m doing when it comes to TLS :/


Lukasa:
😉 No worries: I *do* know about TLS and missed it in code review, so you don't need to beat yourself up too much.

PR #321, opened 2017-05-06
Respect `verify` option in requests adapter

KostyaEsmukov:


test/test_hyper.py
```diff
@@ -1129,6 +1130,29 @@ def test_adapter_accept_client_certificate(self):
...
+
+    def test_adapter_respects_custom_ca_verification(self):
+        a = HTTP20Adapter()
+        conn = a.get_connection(
+            'http2bin.org',
+            80,
+            'http',
+            verify=SERVER_CERT_FILE)
+        assert conn._conn.ssl_context.check_hostname
+        assert conn._conn.ssl_context.verify_mode == ssl.CERT_REQUIRED
```
KostyaEsmukov, line 1155: I would like to ensure here that ssl_context has really loaded the passed CA file, but I can't find a way to check this.

test/test_hyper.py
```diff
@@ -1129,6 +1130,29 @@ def test_adapter_accept_client_certificate(self):
...
+        assert conn1._conn.ssl_context.verify_mode == ssl.CERT_REQUIRED
+
+    def test_adapter_respects_disabled_ca_verification(self):
+        a = HTTP20Adapter()
+        conn = a.get_connection(
+            'http2bin.org',
+            80,
+            'http',
+            verify=False,
+            cert=CLIENT_PEM_FILE)
```
Lukasa, line 1143: Can we validate the client cert is being set?

test/test_hyper.py
Lukasa: Yeah, sadly you can't. =( This is one of the many deficiencies of the OpenSSL interface we have access to.

Lukasa (changes requested):
Thanks, this looks reasonable to me! I have a few small notes inline, but nothing drastic.

test/test_hyper.py
KostyaEsmukov: I don't think so, I can't find it in the `SSLContext` API.
Maybe we could try an integration test instead?

Lukasa: Yeah, that's a good point, I forgot that there was no API for it. Let's just let it ride for now I guess. :grin:

StephanDollberg:
Anything missing on this? Otherwise, I could help getting this over the finish line.

Lukasa (approved):

vikramojha89:
Can you reopen the issue, I am still getting this issue with below script:

```
Accept_header = {'Accept': 'application/xhtml+xml;v=2.0'}

basic_auth = HTTPBasicAuth("User", "Password")

print(type(basic_auth))

s = requests.Session()

s.mount('https://127.0.0.1:443', HTTP20Adapter())
r = s.get('https://127.0.0.1:443/ctrl', auth=basic_auth, headers=Accept_header, verify=False)
print(r.status_code)
```

`ssl.SSLCertVerificationError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self signed certificate (_ssl.c:1076)`

Please let me know if there is issue in my code, or if we have any workaround for now

Issue #304, opened 2017-01-19
HTTPS proxy support


chripede:
Is this planned? I see that proxy support for http is included, but since almost every host only serves HTTP2 on https it seems kind of useless.

Lukasa:
"planned" is a tricky term.

This project is not hugely actively developed at the moment, primarily because my main focus right now is on rewriting urllib3 with an eye towards a HTTP/2 future. This means that I don't have any time to actively develop code for this library. However, I'm happy to code-review patches, including a patch that would add this functionality, so if you or someone else would like to take up the work to do it then I'm happy to work with them.

vfaronov:
@Lukasa 

> my main focus right now is on rewriting urllib3 with an eye towards a HTTP/2 future

Does this mean that Hyper will essentially be obsoleted by urllib3 2.0?

Lukasa:
@vfaronov The answer is "it depends". Eventually, I'd like it to be, yes. But we're a very long way around from that.

Hyper still has an active and valuable role to play in the Python HTTP ecosystem, and I certainly won't prevent people from maintaining and improving Hyper in the long term. Indeed, I'm looking at adding some co-maintainers in future because I don't have a bunch of time to spend on Hyper at the moment.

Certainly, urllib3 v2 will not contain HTTP/2 support: only the hooks necessary for adding it. In the long term we may add it in a future patch, but that's definitely at least 6 months away.

LLCampos:
My understanding is that this issue was solved by #322. Better to close it to avoid confusion?

Lukasa:
Yup. 😁

PR #322, opened 2017-05-06
Implement HTTPS proxy support

KostyaEsmukov:
Related issue: #304

~~This is a very work in progress. It seems to work, but I would like to get your feedback whether I'm on a right track or not.~~

~~The main concern I have is about httplib: is it OK to use it for just sending `CONNECT` method to a proxy?~~ Though [HTTP/2 standard defines][1] CONNECT method as well, I guess it is an overkill to use it for just getting a raw socket from a proxy.

TODO:
- [x] Requests adapter: respect custom request ports when choosing a proxy
- [x] Validate proxy support on h2 w/o tls (h2c) (should we even think about it? I'm not sure)
- [x] ~~raise appropriate IO exceptions to distinguish between connection failure to proxy and a general IO exception~~ (to be done in a follow-up PR) 
- [x] Implement proxy authorization support
- [x] ~~Strip `Proxy-*` headers from response on HTTP proxying~~ (seems unneeded. requests doesn't do it)

[1]: https://tools.ietf.org/html/rfc7540#section-8.3

### DIFF
--- a/hyper/common/exceptions.py
+++ b/hyper/common/exceptions.py
@@ -71,3 +71,22 @@ class MissingCertFile(Exception):
     The certificate file could not be found.
     """
     pass
+
+
+# Create our own ConnectionError.
+try:  # pragma: no cover
+    ConnectionError = ConnectionError
+except NameError:  # pragma: no cover
+    class ConnectionError(Exception):
+        """
+        An error occurred during connection to a host.
+        """
+
+
+class ProxyError(ConnectionError):
+    """
+    An error occurred during connection to a proxy.
+    """
+    def __init__(self, message, response):
+        self.response = response
+        super(ProxyError, self).__init__(message)
--- a/hyper/contrib.py
+++ b/hyper/contrib.py
@@ -9,7 +9,9 @@ try:
     from requests.adapters import HTTPAdapter
     from requests.models import Response
     from requests.structures import CaseInsensitiveDict
-    from requests.utils import get_encoding_from_headers
+    from requests.utils import (
+        get_encoding_from_headers, select_proxy, prepend_scheme_if_needed
+    )
     from requests.cookies import extract_cookies_to_jar
 except ImportError:  # pragma: no cover
     HTTPAdapter = object
@@ -29,7 +31,8 @@ class HTTP20Adapter(HTTPAdapter):
         #: A mapping between HTTP netlocs and ``HTTP20Connection`` objects.
         self.connections = {}
 
-    def get_connection(self, host, port, scheme, cert=None, verify=True):
+    def get_connection(self, host, port, scheme, cert=None, verify=True,
+                       proxy=None):
         """
         Gets an appropriate HTTP/2 connection object based on
         host/port/scheme/cert tuples.
@@ -50,29 +53,51 @@ class HTTP20Adapter(HTTPAdapter):
         elif verify is not True:
             ssl_context = init_context(cert_path=verify, cert=cert)
 
+        if proxy:
+            proxy_headers = self.proxy_headers(proxy)
+            proxy_netloc = urlparse(proxy).netloc
+        else:
+            proxy_headers = None
+            proxy_netloc = None
+
+        # We put proxy headers in the connection_key, because
+        # ``proxy_headers`` method might be overridden, so we can't
+        # rely on proxy headers being the same for the same proxies.
+        proxy_headers_key = (frozenset(proxy_headers.items())
+                             if proxy_headers else None)
+        connection_key = (host, port, scheme, cert, verify,
+                          proxy_netloc, proxy_headers_key)
         try:
-            conn = self.connections[(host, port, scheme, cert, verify)]
+            conn = self.connections[connection_key]
         except KeyError:
             conn = HTTPConnection(
                 host,
                 port,
                 secure=secure,
-                ssl_context=ssl_context)
-            self.connections[(host, port, scheme, cert, verify)] = conn
+                ssl_context=ssl_context,
+                proxy_host=proxy_netloc,
+                proxy_headers=proxy_headers)
+            self.connections[connection_key] = conn
 
         return conn
 
-    def send(self, request, stream=False, cert=None, verify=True, **kwargs):
+    def send(self, request, stream=False, cert=None, verify=True, proxies=None,
+             **kwargs):
         """
         Sends a HTTP message to the server.
         """
+        proxy = select_proxy(request.url, proxies)
+        if proxy:
+            proxy = prepend_scheme_if_needed(proxy, 'http')
+
         parsed = urlparse(request.url)
         conn = self.get_connection(
             parsed.hostname,
             parsed.port,
             parsed.scheme,
             cert=cert,
-            verify=verify)
+            verify=verify,
+            proxy=proxy)
 
         # Build the selector.
         selector = parsed.path
@@ -97,7 +122,7 @@ class HTTP20Adapter(HTTPAdapter):
     def build_response(self, request, resp):
         """
         Builds a Requests' response object.  This emulates most of the logic of
-        the standard fuction but deals with the lack of the ``.headers``
+        the standard function but deals with the lack of the ``.headers``
         property on the HTTP20Response object.
 
         Additionally, this function builds in a number of features that are
--- a/hyper/http11/connection.py
+++ b/hyper/http11/connection.py
@@ -18,9 +18,11 @@ from hyperframe.frame import SettingsFrame
 from .response import HTTP11Response
 from ..tls import wrap_socket, H2C_PROTOCOL
 from ..common.bufsocket import BufferedSocket
-from ..common.exceptions import TLSUpgrade, HTTPUpgrade
+from ..common.exceptions import TLSUpgrade, HTTPUpgrade, ProxyError
 from ..common.headers import HTTPHeaderMap
-from ..common.util import to_bytestring, to_host_port_tuple, HTTPVersion
+from ..common.util import (
+    to_bytestring, to_host_port_tuple, to_native_string, HTTPVersion
+)
 from ..compat import bytes
 
 # We prefer pycohttpparser to the pure-Python interpretation
@@ -34,6 +36,43 @@ log = logging.getLogger(__name__)
 
 BODY_CHUNKED = 1
 BODY_FLAT = 2
+
+
+def _create_tunnel(proxy_host, proxy_port, target_host, target_port,
+                   proxy_headers=None):
+    """
+    Sends CONNECT method to a proxy and returns a socket with established
+    connection to the target.
+
+    :returns: socket
+    """
+    conn = HTTP11Connection(proxy_host, proxy_port)
+    conn.request('CONNECT', '%s:%d' % (target_host, target_port),
+                 headers=proxy_headers)
+
+    resp = conn.get_response()
+    if resp.status != 200:
+        raise ProxyError(
+            "Tunnel connection failed: %d %s" %
+            (resp.status, to_native_string(resp.reason)),
+            response=resp
+        )
+    return conn._sock
+
+
+def _headers_to_http_header_map(headers):
+    # TODO turn this to a classmethod of HTTPHeaderMap
+    headers = headers or {}
+    if not isinstance(headers, HTTPHeaderMap):
+        if isinstance(headers, Mapping):
+            headers = HTTPHeaderMap(headers.items())
+        elif isinstance(headers, Iterable):
+            headers = HTTPHeaderMap(headers)
+        else:
+            raise ValueError(
+                'Header argument must be a dictionary or an iterable'
+            )
+    return headers
 
 
 class HTTP11Connection(object):
@@ -53,14 +92,16 @@ class HTTP11Connection(object):
     :param proxy_host: (optional) The proxy to connect to.  This can be an IP
         address or a host name and may include a port.
     :param proxy_port: (optional) The proxy port to connect to. If not provided
-        and one also isn't provided in the ``proxy`` parameter,
+        and one also isn't provided in the ``proxy_host`` parameter,
         defaults to 8080.
+    :param proxy_headers: (optional) The headers to send to a proxy.
     """
 
     version = HTTPVersion.http11
 
     def __init__(self, host, port=None, secure=None, ssl_context=None,
-                 proxy_host=None, proxy_port=None, **kwargs):
+                 proxy_host=None, proxy_port=None, proxy_headers=None,
+                 **kwargs):
         if port is None:
             self.host, self.port = to_host_port_tuple(host, default_port=80)
         else:
@@ -83,17 +124,21 @@ class HTTP11Connection(object):
         self.ssl_context = ssl_context
         self._sock = None
 
+        # Keep the current request method in order to be able to know
+        # in get_response() what was the request verb.
+        self._current_request_method = None
+
         # Setup proxy details if applicable.
-        if proxy_host:
-            if proxy_port is None:
-                self.proxy_host, self.proxy_port = to_host_port_tuple(
-                    proxy_host, default_port=8080
-                )
-            else:
-                self.proxy_host, self.proxy_port = proxy_host, proxy_port
+        if proxy_host and proxy_port is None:
+            self.proxy_host, self.proxy_port = to_host_port_tuple(
+                proxy_host, default_port=8080
+            )
+        elif proxy_host:
+            self.proxy_host, self.proxy_port = proxy_host, proxy_port
         else:
             self.proxy_host = None
             self.proxy_port = None
+        self.proxy_headers = proxy_headers
 
         #: The size of the in-memory buffer used to store data from the
         #: network. This is used as a performance optimisation. Increase buffer
@@ -113,19 +158,28 @@ class HTTP11Connection(object):
         :returns: Nothing.
         """
         if self._sock is None:
-            if not self.proxy_host:
-                host = self.host
-                port = self.port
-            else:
-                host = self.proxy_host
-                port = self.proxy_port
 
-            sock = socket.create_connection((host, port), 5)
+            if self.proxy_host and self.secure:
+                # Send http CONNECT method to a proxy and acquire the socket
+                sock = _create_tunnel(
+                    self.proxy_host,
+                    self.proxy_port,
+                    self.host,
+                    self.port,
+                    proxy_headers=self.proxy_headers
+                )
+            elif self.proxy_host:
+                # Simple http proxy
+                sock = socket.create_connection(
+                    (self.proxy_host, self.proxy_port),
+                    5
+                )
+            else:
+                sock = socket.create_connection((self.host, self.port), 5)
             proto = None
 
             if self.secure:
-                assert not self.proxy_host, "Proxy with HTTPS not supported."
-                sock, proto = wrap_socket(sock, host, self.ssl_context)
+                sock, proto = wrap_socket(sock, self.host, self.ssl_context)
 
             log.debug("Selected protocol: %s", proto)
             sock = BufferedSocket(sock, self.network_buffer_size)
@@ -154,25 +208,29 @@ class HTTP11Connection(object):
         :returns: Nothing.
         """
 
-        headers = headers or {}
-
         method = to_bytestring(method)
+        is_connect_method = b'CONNECT' == method.upper()
+        self._current_request_method = method
+
+        if self.proxy_host and not self.secure:
+            # As per https://tools.ietf.org/html/rfc2068#section-5.1.2:
+            # The absoluteURI form is required when the request is being made
+            # to a proxy.
+            url = self._absolute_http_url(url)
         url = to_bytestring(url)
 
-        if not isinstance(headers, HTTPHeaderMap):
-            if isinstance(headers, Mapping):
-                headers = HTTPHeaderMap(headers.items())
-            elif isinstance(headers, Iterable):
-                headers = HTTPHeaderMap(headers)
-            else:
-                raise ValueError(
-                    'Header argument must be a dictionary or an iterable'
-                )
+        headers = _headers_to_http_header_map(headers)
+
+        # Append proxy headers.
+        if self.proxy_host and not self.secure:
+            headers.update(
+                _headers_to_http_header_map(self.proxy_headers).items()
+            )
 
         if self._sock is None:
             self.connect()
 
-        if self._send_http_upgrade:
+        if not is_connect_method and self._send_http_upgrade:
             self._add_upgrade_headers(headers)
             self._send_http_upgrade = False
 
@@ -180,7 +238,7 @@ class HTTP11Connection(object):
         if body:
             body_type = self._add_body_headers(headers, body)
 
-        if b'host' not in headers:
+        if not is_connect_method and b'host' not in headers:
             headers[b'host'] = self.host
 
         # Begin by emitting the header block.
@@ -192,6 +250,10 @@ class HTTP11Connection(object):
 
         return
 
+    def _absolute_http_url(self, url):
+        port_part = ':%d' % self.port if self.port != 80 else ''
+        return 'http://%s%s%s' % (self.host, port_part, url)
+
     def get_response(self):
         """
         Returns a response object.
@@ -199,6 +261,9 @@ class HTTP11Connection(object):
         This is an early beta, so the response object is pretty stupid. That's
         ok, we'll fix it later.
         """
+        method = self._current_request_method
+        self._current_request_method = None
+
         headers = HTTPHeaderMap()
 
         response = None
@@ -228,7 +293,8 @@ class HTTP11Connection(object):
             response.msg.tobytes(),
             headers,
             self._sock,
-            self
+            self,
+            method
         )
 
     def _send_headers(self, method, url, headers):
--- a/hyper/http11/response.py
+++ b/hyper/http11/response.py
@@ -27,7 +27,8 @@ class HTTP11Response(object):
 
     version = HTTPVersion.http11
 
-    def __init__(self, code, reason, headers, sock, connection=None):
+    def __init__(self, code, reason, headers, sock, connection=None,
+                 request_method=None):
         #: The reason phrase returned by the server.
         self.reason = reason
 
@@ -62,11 +63,19 @@ class HTTP11Response(object):
             b'chunked' in self.headers.get(b'transfer-encoding', [])
         )
 
-        # One of the following must be true: we must expect that the connection
-        # will be closed following the body, or that a content-length was sent,
-        # or that we're getting a chunked response.
-        # FIXME: Remove naked assert, replace with something better.
-        assert self._expect_close or self._length is not None or self._chunked
+        # When content-length is absent and response is not chunked,
+        # body length is determined by connection closure.
+        # https://tools.ietf.org/html/rfc7230#section-3.3.3
+        if self._length is None and not self._chunked:
+            # 200 response to a CONNECT request means that proxy has connected
+            # to the target host and it will start forwarding everything sent
+            # from the either side. Thus we must not try to read body of this
+            # response. Socket of current connection will be taken over by
+            # the code that has sent a CONNECT request.
+            if not (request_method is not None and
+                    b'CONNECT' == request_method.upper() and
+                    code == 200):
+                self._expect_close = True
 
         # This object is used for decompressing gzipped request bodies. Right
         # now we only support gzip because that's all the RFC mandates of us.
--- a/hyper/http20/connection.py
+++ b/hyper/http20/connection.py
@@ -18,6 +18,7 @@ from ..common.util import (
     to_host_port_tuple, to_native_string, to_bytestring, HTTPVersion
 )
 from ..compat import unicode, bytes
+from ..http11.connection import _create_tunnel
 from .stream import Stream
 from .response import HTTP20Response, HTTP20Push
 from .window import FlowControlManager
@@ -29,6 +30,7 @@ import logging
 import socket
 import time
 import threading
+import itertools
 
 log = logging.getLogger(__name__)
 
@@ -90,15 +92,17 @@ class HTTP20Connection(object):
     :param proxy_host: (optional) The proxy to connect to.  This can be an IP
         address or a host name and may include a port.
     :param proxy_port: (optional) The proxy port to connect to. If not provided
-        and one also isn't provided in the ``proxy`` parameter, defaults to
-        8080.
+        and one also isn't provided in the ``proxy_host`` parameter, defaults
+        to 8080.
+    :param proxy_headers: (optional) The headers to send to a proxy.
     """
 
     version = HTTPVersion.http20
 
     def __init__(self, host, port=None, secure=None, window_manager=None,
                  enable_push=False, ssl_context=None, proxy_host=None,
-                 proxy_port=None, force_proto=None, **kwargs):
+                 proxy_port=None, force_proto=None, proxy_headers=None,
+                 **kwargs):
         """
         Creates an HTTP/2 connection to a specific server.
         """
@@ -118,16 +122,16 @@ class HTTP20Connection(object):
         self.ssl_context = ssl_context
 
         # Setup proxy details if applicable.
-        if proxy_host:
-            if proxy_port is None:
-                self.proxy_host, self.proxy_port = to_host_port_tuple(
-                    proxy_host, default_port=8080
-                )
-            else:
-                self.proxy_host, self.proxy_port = proxy_host, proxy_port
+        if proxy_host and proxy_port is None:
+            self.proxy_host, self.proxy_port = to_host_port_tuple(
+                proxy_host, default_port=8080
+            )
+        elif proxy_host:
+            self.proxy_host, self.proxy_port = proxy_host, proxy_port
         else:
             self.proxy_host = None
             self.proxy_port = None
+        self.proxy_headers = proxy_headers
 
         #: The size of the in-memory buffer used to store data from the
         #: network. This is used as a performance optimisation. Increase buffer
@@ -272,10 +276,18 @@ class HTTP20Connection(object):
         # being sent in the wrong order, which can lead to the out-of-order
         # messages with lower stream IDs being closed prematurely.
         with self._write_lock:
+            # Unlike HTTP/1.1, HTTP/2 (according to RFC 7540) doesn't require
+            # to use absolute URI when proxying.
+
             stream_id = self.putrequest(method, url)
 
             default_headers = (':method', ':scheme', ':authority', ':path')
-            for name, value in headers.items():
+            all_headers = headers.items()
+            if self.proxy_host and not self.secure:
+                proxy_headers = self.proxy_headers or {}
+                all_headers = itertools.chain(all_headers,
+                                              proxy_headers.items())
+            for name, value in all_headers:
                 is_default = to_native_string(name) in default_headers
                 self.putheader(name, value, stream_id, replace=is_default)
 
@@ -358,18 +370,25 @@ class HTTP20Connection(object):
             if self._sock is not None:
                 return
 
-            if not self.proxy_host:
-                host = self.host
-                port = self.port
+            if self.proxy_host and self.secure:
+                # Send http CONNECT method to a proxy and acquire the socket
+                sock = _create_tunnel(
+                    self.proxy_host,
+                    self.proxy_port,
+                    self.host,
+                    self.port,
+                    proxy_headers=self.proxy_headers
+                )
+            elif self.proxy_host:
+                # Simple http proxy
+                sock = socket.create_connection(
+                    (self.proxy_host, self.proxy_port)
+                )
             else:
-                host = self.proxy_host
-                port = self.proxy_port
-
-            sock = socket.create_connection((host, port))
+                sock = socket.create_connection((self.host, self.port))
 
             if self.secure:
-                assert not self.proxy_host, "Proxy with HTTPS not supported."
-                sock, proto = wrap_socket(sock, host, self.ssl_context,
+                sock, proto = wrap_socket(sock, self.host, self.ssl_context,
                                           force_proto=self.force_proto)
             else:
                 proto = H2C_PROTOCOL
--- a/test/server.py
+++ b/test/server.py
@@ -15,6 +15,7 @@ ingenuity and about a million beers. The license is available in NOTICES.
 import threading
 import socket
 import sys
+from enum import Enum
 
 from hyper import HTTP20Connection
 from hyper.compat import ssl
@@ -25,6 +26,23 @@ from hpack.huffman_constants import (
     REQUEST_CODES, REQUEST_CODES_LENGTH
 )
 from hyper.tls import NPN_PROTOCOL
+
+
+class SocketSecuritySetting(Enum):
+    """
+    Server socket TLS wrapping strategy:
+
+    SECURE - automatically wrap socket
+    INSECURE - never wrap
+    SECURE_NO_AUTO_WRAP - init context, but socket must be wrapped manually
+
+    The values are needed to be able to convert ``secure`` boolean flag of
+    a client to a member of this enum:
+    ``socket_security = SocketSecuritySetting(secure)``
+    """
+    SECURE = True
+    INSECURE = False
+    SECURE_NO_AUTO_WRAP = 'NO_AUTO_WRAP'
 
 
 class SocketServerThread(threading.Thread):
@@ -42,16 +60,17 @@ class SocketServerThread(threading.Thread):
                  host='localhost',
                  ready_event=None,
                  h2=True,
-                 secure=True):
+                 socket_security=SocketSecuritySetting.SECURE):
         threading.Thread.__init__(self)
 
         self.socket_handler = socket_handler
         self.host = host
-        self.secure = secure
+        self.socket_security = socket_security
         self.ready_event = ready_event
         self.daemon = True
 
-        if self.secure:
+        if self.socket_security in (SocketSecuritySetting.SECURE,
+                                    SocketSecuritySetting.SECURE_NO_AUTO_WRAP):
             self.cxt = ssl.SSLContext(ssl.PROTOCOL_SSLv23)
             if ssl.HAS_NPN and h2:
                 self.cxt.set_npn_protocols([NPN_PROTOCOL])
@@ -63,8 +82,8 @@ class SocketServerThread(threading.Thread):
         if sys.platform != 'win32':
             sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
 
-        if self.secure:
-            sock = self.cxt.wrap_socket(sock, server_side=True)
+        if self.socket_security == SocketSecuritySetting.SECURE:
+            sock = self.wrap_socket(sock)
         sock.bind((self.host, 0))
         self.port = sock.getsockname()[1]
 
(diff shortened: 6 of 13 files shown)


test/server.py
```diff
@@ -78,7 +80,7 @@ def _start_server(self):
         sock.close()
 
     def _wrap_socket(self, sock):
-        raise NotImplementedError()
+        return self.cxt.wrap_socket(sock, server_side=True)
```
KostyaEsmukov: Is it OK? What was the purpose of this method? I couldn't find any references to it

test/test_integration.py
```diff
@@ -672,6 +672,68 @@ def socket_handler(listener):
...
+
+            # Read the CONNECT reader
+            connect_data = b''
+            while not connect_data.endswith(b'\r\n\r\n'):
+                connect_data += sock.recv(65535)
+
+            sock.send(b'HTTP/1.0 200 Connection established\r\n\r\n')
+
+            # todo make this less ugly?
+            sock = self.server_thread._wrap_socket(sock)
```
KostyaEsmukov: I feel really bad about this. There must be a better way to wrap server socket. Any ideas?

Lukasa (changes requested):
So, the first note is: no, don't use httplib. =) The codebase has a HTTP/1.1 stack, you should use that. :smile:

hyper/common/exceptions.py
```diff
@@ -71,3 +71,10 @@ class MissingCertFile(Exception):
     The certificate file could not be found.
     """
     pass
+
+
+class ProxyError(Exception):
```
KostyaEsmukov: I feel that it should extend a more generic `ConnectionError` exception, but there's no such common exception. But there's one in `hyper.http20.exceptions.ConnectionError`. Maybe we could introduce it here?

hyper/http20/connection.py
```diff
@@ -18,6 +18,7 @@
     to_host_port_tuple, to_native_string, to_bytestring, HTTPVersion
 )
 from ..compat import unicode, bytes
+from ..http11.connection import HTTP11Connection
```
KostyaEsmukov: Is it OK to explicitly depend on `HTTP11Connection` here?

hyper/common/exceptions.py
Lukasa: I'm happy to introduce a common `ConnectionError` that the HTTP/2 one subclasses.

hyper/http20/connection.py
Lukasa: Yes, I think so.

hyper/contrib.py
```diff
@@ -29,11 +31,18 @@ def __init__(self, *args, **kwargs):
...
         self.connections = {}
 
-    def get_connection(self, host, port, scheme, cert=None):
+    def get_connection(self, host, port, scheme, cert=None, proxies=None):
         """
         Gets an appropriate HTTP/2 connection object based on
         host/port/scheme/cert tuples.
         """
+        # TODO take custom request ports into account
+        proxy = select_proxy("%s://%s" % (scheme, host), proxies)
```
Lukasa: We should probably do this at a higher scope so that we can pass `request.url` instead of rebuilding it ourselves. That's what the Requests version of this code does.

hyper/http11/connection.py
```diff
@@ -137,6 +144,33 @@ def connect(self):
 
         return
 
+    @classmethod
```
Lukasa: So this is weird, I know, but I kind of hate classmethods for this use case. IMO, classmethods are for custom constructors, and nothing else. This should just be a function that uses `HTTP11Connection` directly, I think.

hyper/http11/connection.py
```diff
@@ -113,19 +115,24 @@ def connect(self):
...
-            else:
-                host = self.proxy_host
-                port = self.proxy_port
 
-            sock = socket.create_connection((host, port), 5)
+            if self.proxy_host and self.secure:
+                # http CONNECT proxy
+                sock = HTTP11Connection._create_tunnel(
+                    self.proxy_host, self.proxy_port, self.host, self.port)
+            else:
```
Lukasa: This branch is probably simpler as two `elif` branches that just consist of a one-line call to `socket.create_connection`.

hyper/http11/connection.py
```diff
@@ -137,6 +144,33 @@ def connect(self):
...
+    def _create_tunnel(cls, proxy_host, proxy_port, target_host, target_port):
+        """
+        Sends CONNECT method to a proxy and returns a socket with established
+        connection to the target.
+
+        :returns: socket
+        """
+        conn = cls(proxy_host, proxy_port)
+        conn.connect()
+        conn._send_headers(to_bytestring('CONNECT'),
```
Lukasa: As this is a constant, this can just be `b'CONNECT'`. We only need `to_bytestring` to allow string formatting operations.

hyper/http11/connection.py
```diff
@@ -137,6 +144,33 @@ def connect(self):
...
+                           to_bytestring('%s:%d' % (target_host, target_port)),
+                           # TODO proxy-authorization headers
+                           headers=HTTPHeaderMap())
+
+        response = None
+        while response is None:
+            # 'encourage' the socket to receive data.
+            conn._sock.fill()
+            response = conn.parser.parse_response(conn._sock.buffer)
```
Lukasa: Is there any reason we can't just actually use the regular request/response logic, and just not preload any data?

hyper/http11/connection.py
```diff
@@ -157,6 +191,10 @@ def request(self, method, url, body=None, headers=None):
         headers = headers or {}
 
         method = to_bytestring(method)
+
+        if self.proxy_host and not self.secure:
+            port_part = ':%d' % self.port if self.port != 80 else ''
+            url = 'http://%s%s%s' % (self.host, port_part, url)
```
Lukasa: It'd be nice to pull this out to a helper function.

hyper/http20/connection.py
Lukasa: Yup, totally fine.

hyper/http20/connection.py
```diff
@@ -272,6 +273,8 @@ def request(self, method, url, body=None, headers=None):
         # being sent in the wrong order, which can lead to the out-of-order
         # messages with lower stream IDs being closed prematurely.
         with self._write_lock:
+            # TODO should we send an absolute url as in http11 ...
+            # ... when (self.proxy_host and not self.secure)?
```
Lukasa: No, RFC 7540 doesn't need you to do it. The `:authority` header field is all that is needed.

hyper/http20/connection.py
```diff
@@ -358,18 +361,22 @@ def connect(self):
...
-
-            sock = socket.create_connection((host, port))
+                if self.proxy_host:
+                    # simple http proxy
+                    host = self.proxy_host
+                    port = self.proxy_port
+                else:
+                    host = self.host
+                    port = self.port
+                sock = socket.create_connection((host, port))
```
Lukasa: All notes in HTTP/1.1 logic apply here.

test/server.py
```diff
@@ -42,12 +42,14 @@ def __init__(self,
                  host='localhost',
                  ready_event=None,
                  h2=True,
-                 secure=True):
+                 secure=True,
+                 secure_auto_wrap_socket=True):
```
Lukasa: It'd be nicer, rather than adding boolean fields that only make sense in the presence of other boolean fields, to turn `secure` into a field that accepts an `enum`, and have that `enum` include the relevant options.

test/server.py
Lukasa: Good question, I have no idea.

test/server.py
```diff
@@ -89,11 +91,12 @@ class SocketLevelTest(object):
     A test-class that defines a few helper methods for running socket-level
     tests.
     """
-    def set_up(self, secure=True, proxy=False):
+    def set_up(self, secure=True, proxy=False, secure_auto_wrap_socket=True):
```
Lukasa: Same note here.

test/test_integration.py
```diff
@@ -672,6 +673,98 @@ def socket_handler(listener):
...
+        self.set_up(secure=True, secure_auto_wrap_socket=False, proxy=True)
+
+        data = []
+        req_event = threading.Event()
+        recv_event = threading.Event()
+
+        def socket_handler(listener):
+            sock = listener.accept()[0]
+
+            # Read the CONNECT reader
```
Lukasa: s/reader/header/

test/test_integration.py
Lukasa: I think this is fine, TBH. Maybe promote `_wrap_socket` to `wrap_socket`, or even pull that `wrap_socket` method off the server thread and make it available as a test helper function instead.

test/test_integration.py
```diff
@@ -672,6 +673,98 @@ def socket_handler(listener):
...
+
+    def test_failing_proxy_tunnel(self):
+        self.set_up(secure=True, secure_auto_wrap_socket=False, proxy=True)
+
+        recv_event = threading.Event()
+
+        def socket_handler(listener):
+            sock = listener.accept()[0]
+
+            # Read the CONNECT reader
```
Lukasa: s/reader/header/

test/test_integration.py
```diff
@@ -1229,3 +1322,57 @@ def socket_handler(listener):
...
+
+    def test_adapter_uses_proxies(self):
+        self.set_up(secure=True, secure_auto_wrap_socket=False, proxy=True)
+
+        send_event = threading.Event()
+
+        def socket_handler(listener):
+            sock = listener.accept()[0]
+
+            # Read the CONNECT reader
```
Lukasa: s/reader/helper/

test/test_integration_http11.py
```diff
@@ -163,6 +163,60 @@ def socket_handler(listener):
...
 
+    def test_secure_proxy_request_response(self):
+        self.set_up(secure=True, secure_auto_wrap_socket=False, proxy=True)
+
+        send_event = threading.Event()
+
+        def socket_handler(listener):
+            sock = listener.accept()[0]
+
+            # Read the CONNECT reader
```
Lukasa: s/reader/header/

Lukasa (changes requested):
Cool, so this looks like a fairly good start, well done! I have left a bunch of code review feedback, but with a largish patch like this that's always going to happen, especially when you are doing the right thing and putting it forward for feedback early. Well done so far, I'm looking forward to seeing how this moves forward!

hyper/http11/connection.py
KostyaEsmukov: The reason I've made it a classmethod is to ensure that the `HTTP11Connection` instance, created to connect to a proxy, is not usable/alterable anymore, because its socket is moved to an original connection instance, which picks it up.

Maybe a staticmethod will be better? I'm afraid that an instance method doesn't incapsulate this fragile socket transition properly

hyper/http11/connection.py
KostyaEsmukov: Yes, the reason is that `HTTP11Response` `assert`s on presence of some headers like `Connection`/`Content-length` which might be absent in the proxy response to a CONNECT. 

Maybe I should add a comment here explaining this. Or should we adapt the `HTTP11Response` to accept responses without these headers?

test/test_integration.py
KostyaEsmukov: Aww, what a shameful copy-pasted typo 😃

BTW, CONNECT is a method, [according to RFC 7231 HTTP/1.1](https://tools.ietf.org/html/rfc7231#section-4.3.6)
I will look through my comments again, thanks!

test/server.py
KostyaEsmukov: Yes, I thought about this, but it would have required to replace a lot of True/False values for current tests, producing a meaninglessly (at this moment) large diff.

However, I think I've got a better idea how to make it work without this new boolean flag.  Patch is on the way 💪

Update:
Nah, it didn't work out 😄. I wanted to use `secure` as `secure_auto_wrap_socket` alone, but I forgot that `SocketLevelTest.get_connection` uses `secure` as a flag for a client connection. I've introduced enum.

hyper/common/exceptions.py
KostyaEsmukov: There is a `hyper.common.exceptions.SocketError`, which is not referenced anywhere in the project. It's very similar to the `ConnectionError` I've just introduced, I guess.

I suggest to leave them both here for now, but I would like to work on exceptions structure in another PR after this one

KostyaEsmukov:
Thank you for your feedback, I appreciate it!
I guess I have addressed all of the notes.
I'm going to implement the last two points of my TODO today and then we may start polishing it for a final state 😊

hyper/http11/connection.py
Lukasa: Sorry, I should have been clearer. I wasn't suggesting an **instance method**: I was suggesting a **function**. Don't hang it off a class at all, it doesn't need one.

hyper/http11/connection.py
Lukasa: I think we should probably adapt them to accept responses without those headers.

test/test_integration.py
Lukasa: It is a method, yes, but in this case we're reading the whole request header. :wink:

Lukasa:
Cool, in that case I'll hold off reviewing until you deal with your last few TODOs. Try to reduce the amount of back-and-forth we do. :smile:

KostyaEsmukov:
Eww, this diff is already big! I guess I've done with this feature. There's more I want to do in hyper, but lets hold it for another PR(s).
As for now proxy support is complete IMO. Waiting for a review 😊

hyper/contrib.py
```diff
@@ -43,28 +48,40 @@ def get_connection(self, host, port, scheme, cert=None):
         if cert is not None:
             ssl_context = init_context(cert=cert)
 
+        connection_key = (host, port, scheme, cert, proxy)
```
KostyaEsmukov: Calling `get_connection` with the same arguments but different `proxy_headers` would reuse already created connection, so the new `proxy_headers` will be ignored.

Our `proxy_headers` are generated from proxy url, so we should be fine, but this is not strictly guaranteed in this code.

Maybe we should move in the `select_proxy` logic here, thus generating `proxy_headers` inside this function?

Lukasa: Any reason not to just make `proxy_headers` part of the tuple?

hyper/contrib.py
```diff
@@ -157,3 +174,16 @@ def getheaders(self, name):
...
+        Returns a dictionary of the headers to add to any request sent
+        through a proxy.
+        """
+        headers = {}
+        username, password = get_auth_from_url(proxy)
+
+        if username:
+            headers['Proxy-Authorization'] = _basic_auth_str(username,
+                                                             password)
+        return headers
```
Lukasa: Do we need this code? We inherit from the `HTTPAdapter`, which already provides this method. Why can't we just call it?

hyper/http11/connection.py
```diff
@@ -396,3 +427,23 @@ def __enter__(self):
     def __exit__(self, type, value, tb):
         self.close()
         return False  # Never swallow exceptions.
+
+
+def _create_tunnel(proxy_host, proxy_port, target_host, target_port,
```
Lukasa: As a minor coding style note, it's better to put functions like these towards the *top* of the file rather than towards the bottom.

hyper/http11/connection.py
```diff
@@ -113,19 +121,22 @@ def connect(self):
...
-            else:
-                host = self.proxy_host
-                port = self.proxy_port
 
-            sock = socket.create_connection((host, port), 5)
+            if self.proxy_host and self.secure:
+                # Send http CONNECT method to a proxy and acquire the socket
+                sock = _create_tunnel(self.proxy_host, self.proxy_port,
+                                      self.host, self.port,
+                                      proxy_headers=self.proxy_headers)
```
Lukasa: The standard style in this codebase for breaking the line would be like this:

```python
sock = _create_tunnel(
    self.proxy_host,
    self.proxy_port,
    self.host,
    self.port,
    proxy_headers=self.proxy_headers
)
```

hyper/http11/connection.py
```diff
@@ -113,19 +121,22 @@ def connect(self):
...
+            elif self.proxy_host:
+                # Simple http proxy
+                sock = socket.create_connection((self.proxy_host,
+                                                 self.proxy_port), 5)
```
Lukasa: Same multiline note.

hyper/http11/connection.py
```diff
@@ -192,6 +204,23 @@ def request(self, method, url, body=None, headers=None):
...
+        if not isinstance(headers, HTTPHeaderMap):
+            if isinstance(headers, Mapping):
+                headers = HTTPHeaderMap(headers.items())
+            elif isinstance(headers, Iterable):
+                headers = HTTPHeaderMap(headers)
+            else:
+                raise ValueError(
+                    'Header argument must be a dictionary or an iterable'
+                )
+        return headers
```
Lukasa: This "method" does not use the word `self` anywhere inside it, and so is not a method at all: it's a function pretending to be one. ;)

hyper/http11/connection.py
```diff
@@ -80,20 +84,24 @@ def __init__(self, host, port=None, secure=None, ssl_context=None,
         self._send_http_upgrade = not self.secure
         self._enable_push = kwargs.get('enable_push')
 
+        # CONNECT request (to create a tunnel via proxy) has a few special
+        # cases which needs to be processed differently.
+        self._is_connect_request = False
```
Lukasa: So, I don't think this feels right. A connection is not a request: a request is just one of many things that can happen on a connection. In this case, all of the places it's used except for the response have access to the request verb, so they can just check what the request verb is. The better solution is to *tell* the response what the request verb was, which is good practice anyway.

hyper/http11/connection.py
```diff
@@ -218,7 +247,8 @@ def get_response(self):
         # https://github.com/Lukasa/hyper/issues/312.
         # Connection options are case-insensitive, while upgrade tokens are
         # case-sensitive: https://github.com/httpwg/http11bis/issues/8.
-        if (response.status == 101 and
+        if (not self._is_connect_request and
```
Lukasa: This shouldn't be necessary: as we don't send `Upgrade` headers on a `CONNECT` request, we're not going to get a `101 Switching Protocols` response back. The standard logic is fine.

hyper/http20/connection.py
```diff
@@ -272,13 +275,23 @@ def request(self, method, url, body=None, headers=None):
...
             for name, value in headers.items():
                 is_default = to_native_string(name) in default_headers
                 self.putheader(name, value, stream_id, replace=is_default)
 
+            # Append proxy headers.
+            if self.proxy_host and not self.secure:
+                proxy_headers = self.proxy_headers or {}
+                for name, value in proxy_headers.items():
+                    is_default = to_native_string(name) in default_headers
+                    self.putheader(name, value, stream_id, replace=is_default)
```
Lukasa: It would be better to try to be a bit clever here. We can avoid duplicating the loop above by doing this:

```python
all_headers = headers.items()
if self.proxy_host and not self.secure:
    proxy_headers = self.proxy_headers or {}
    all_headers = itertools.chain(all_headers, proxy_headers.items())

for name, value in all_headers:
```

That'll keep a single loop, and generally seem a lot cleaner.

test/server.py
```diff
@@ -27,6 +28,12 @@
 from hyper.tls import NPN_PROTOCOL
 
 
+class SocketSecure(Enum):
```
Lukasa: I don't love this name. Maybe `SocketSecuritySetting`?

test/server.py
```diff
@@ -27,6 +28,12 @@
...
+    SECURE = True
+    INSECURE = False
+    SECURE_NO_AUTO_WRAP = 'NO_AUTO_WRAP'
```
Lukasa: We don't appear to need the values to be set.

Lukasa (changes requested):
Cool, got some more notes. I didn't get around to reviewing the tests yet, but there's some feedback in the code itself so let's worry about that first. :smile:

hyper/contrib.py
KostyaEsmukov: yes, dict is unhashable

hyper/contrib.py
KostyaEsmukov: Good point! I must have missed that

hyper/http11/connection.py
KostyaEsmukov: OK, I think I've got this point, will try this out tomorrow

hyper/http11/connection.py
KostyaEsmukov: So should it be moved to the top of the file as a function rather than a method of the class? Or maybe even as a classmethod (?) of `HTTPHeaderMap` itself?

BTW, http11 connection accepts headers as Iterable, but http2 connection assumes that headers is always a Mapping (having `.items()` method).

I guess, for clarity, either http20 should use the same function as this one or http11 should also assume that headers is a Mapping.

test/server.py
KostyaEsmukov: Could you please elaborate on this?
I think we actually need the values to be able to convert boolean `secure` to an instance of this enum

hyper/contrib.py
Lukasa: Yeah, but frozenset of tuple isn't. You could write a transform that does `proxy_headers_key = frozenset((k, v) for k, v in proxy_headers.items())`.

hyper/http11/connection.py
Lukasa: For now, top of the file as a function. I'd consider a classmethod of `HTTPHeaderMap`, but let's save that for a follow-up PR rather than shoving it in this one, as both of us are already a bit nervous about the size of the diff. :wink:

Agreed that the two are inconsistent, but let's leave them inconsistent for now and fix them up later with a useful `HTTPHeaderMap` classmethod in a follow-up PR.

test/server.py
Lukasa: Hrm. I suppose? It feels pretty weird though, let's put a comment in here to say that's why we're doing it.

hyper/http11/connection.py
```diff
@@ -36,6 +38,25 @@
...
+    :returns: socket
+    """
+    conn = HTTP11Connection(proxy_host, proxy_port)
+    conn.request('CONNECT', '%s:%d' % (target_host, target_port),
+                 headers=proxy_headers)
+
+    resp = conn.get_response()
+    if resp.status != 200:
+        raise ProxyError("Tunnel connection failed: %d %s" % (
+            resp.status, to_native_string(resp.reason)))
```
Lukasa: It might be nicer to actually return the response object in this case as well, so that users can grab it to read any response body or headers that might be there.

hyper/http11/connection.py
```diff
@@ -83,17 +106,21 @@ def __init__(self, host, port=None, secure=None, ssl_context=None,
         self.ssl_context = ssl_context
         self._sock = None
 
+        # Keep request methods in a queue in order to be able to know
+        # in get_response() what was the request verb.
+        self._request_methods = []
```
Lukasa: Hrm, this doesn't make me much happier, as it turns out. 😞 

I think the thing to do is to keep track of *only* the currently outstanding request method. We don't allow pipelining, really, so we shouldn't have a list: just a single value. We can then use that in `get_response`. Does that sound more sensible?

Lukasa (changes requested):

hyper/http11/connection.py
KostyaEsmukov: This is intended to be a private API, so _we_ are the users of it :)
We don't seem to need this response as for now. Maybe we could leave it as is? 🙂

The problem is to differ socket from response on consumer side, which seems like an unnecessary complication as for now, I think

hyper/http11/connection.py
KostyaEsmukov: Yeah, I was thinking about pipelining while considering a list here.
OK, I will turn this to a single value

hyper/contrib.py
KostyaEsmukov: Please let me know what do you think about https://github.com/Lukasa/hyper/pull/322/commits/bbf59ff8597f8be68c3fe43093762966e826050c

I don't like storing the whole dict as a key. It seems to bloat the key without much benefit

hyper/http11/connection.py
Lukasa: It's a private API in the sense that no-one else calls it, but presumably we allow this exception to bubble up, right?

hyper/contrib.py
Lukasa: I don't think it resolves the core issue. If you are setting headers for the proxy (e.g. proxy-authorization) then we absolutely must respect that. We need to keep track of them. In 99%+ of cases we will never store anything other than one set of proxy headers and a `None` value, which is pretty minimal IMO.

hyper/http11/connection.py
KostyaEsmukov: Right. Do you mean that we should add response as exception attribute?

Lukasa: Yup.

hyper/contrib.py
KostyaEsmukov: Well, OK, will do, but still, I think storing proxy headers in this key is redundant, because they are generated from a proxy which is already a part of this key

Lukasa: They needn't be. Requests has a `proxy_headers` method expressly to allow it to be overridden, in case users want to provide other forms of proxy header handling.

KostyaEsmukov: Alright, that makes sense!
Should I revert that commit, or just add the dict to the key?

Lukasa: At this point we have enough commits that you don't need to worry about reverting: we'll need to squash before we merge to clean them up anyway. :smile: So just make the change directly to get the code looking the way you want it. :grin:

hyper/common/exceptions.py
```diff
@@ -71,3 +71,22 @@ class MissingCertFile(Exception):
...
+        """
+        An error occurred during connection to a host.
+        """
+
+
+class ProxyError(ConnectionError):
+    """
+    An error occurred during connection to a proxy.
+    """
+    def __init__(self, *args, **kwargs):
```
Lukasa: Cool, this is good, but let's save ourselves a `kwargs.pop` and write the declaration as `def __init__(self, *args, response=None, **kwargs):`. That way we save the dict pop, make it clear that we are really expecting a response argument, and just generally make everything a bit nicer. :smile:

hyper/contrib.py
```diff
@@ -43,28 +45,48 @@ def get_connection(self, host, port, scheme, cert=None):
...
+        else:
+            proxy_headers = None
+            proxy_netloc = None
+
+        # We put proxy headers in the connection_key, because
+        # ``proxy_headers`` method might be overridden, so we can't
+        # rely on proxy headers being the same for the same proxies.
+        proxy_headers_key = (frozenset(proxy_headers.items())
+                             if proxy_headers else None)
+        connection_key = (host, port, scheme, cert, proxy, proxy_headers_key)
```
Lukasa: Let me ask: is there a reason to key on `proxy`, instead of `proxy_netloc`?

hyper/http11/connection.py
```diff
@@ -83,17 +124,21 @@ def __init__(self, host, port=None, secure=None, ssl_context=None,
...
-                self.proxy_host, self.proxy_port = to_host_port_tuple(
-                    proxy_host, default_port=8080
-                )
-            else:
-                self.proxy_host, self.proxy_port = proxy_host, proxy_port
+        if proxy_host and proxy_port is None:
+            self.proxy_host, self.proxy_port = to_host_port_tuple(
+                proxy_host, default_port=8080
+            )
+        elif proxy_host and proxy_port is not None:
```
Lukasa: We can condense this: this can just be `elif proxy_host`, as definitionally if we didn't enter the block above but `proxy_host` is true, `proxy_port` must not be `None`.

hyper/http11/connection.py
```diff
@@ -192,13 +250,18 @@ def request(self, method, url, body=None, headers=None):
...
+        return 'http://%s%s%s' % (self.host, port_part, url)
+
     def get_response(self):
         """
         Returns a response object.
 
         This is an early beta, so the response object is pretty stupid. That's
         ok, we'll fix it later.
         """
+        method = self._current_request_method
```
Lukasa, line 264: It'd be nice to `None` out `self._current_request_method` here as well.

hyper/http11/response.py
```diff
@@ -48,7 +49,8 @@ def __init__(self, code, reason, headers, sock, connection=None):
         # bother checking for content-length, we just keep reading until
         # we no longer can.
         self._expect_close = False
-        if b'close' in self.headers.get(b'connection', []):
+        if (b'close' in self.headers.get(b'connection', []) or
+                b'close' in self.headers.get(b'proxy-connection', [])):
             self._expect_close = True
```
Lukasa: Do we want `self._expect_close` to be true for `CONNECT` requests? Probably not, right?

hyper/http20/connection.py
```diff
@@ -118,16 +122,16 @@ def __init__(self, host, port=None, secure=None, window_manager=None,
...
-                self.proxy_host, self.proxy_port = to_host_port_tuple(
-                    proxy_host, default_port=8080
-                )
-            else:
-                self.proxy_host, self.proxy_port = proxy_host, proxy_port
+        if proxy_host and proxy_port is None:
+            self.proxy_host, self.proxy_port = to_host_port_tuple(
+                proxy_host, default_port=8080
+            )
+        elif proxy_host and proxy_port is not None:
```
Lukasa: Same note here about the `proxy_port` check.

Lukasa (changes requested):
Cool, this is coming along really nicely! The number of notes I have each time we go around is shrinking, with is definitely a good sign. I think we're getting very close to this being ready to go.

hyper/common/exceptions.py
KostyaEsmukov: Yeah, that would have been nicer, but Python 2.7 doesn't support this syntax :(
It's in https://www.python.org/dev/peps/pep-3102/

hyper/contrib.py
KostyaEsmukov: Good point! No, there's not, since we include proxy headers in the key

hyper/http11/response.py
KostyaEsmukov: When a proxy responds with 200 to a CONNECT request, it doesn't include a `proxy-connection` header at all (at least I haven't seen that).
But when it responds with >=400, it does include a `proxy-connection: close` header, which should be respected here.

So I think there's no need to make any exception for a CONNECT request here.

UPDATE:
There's a test checking that socket is actially closed in this case. It's `test_integration_http11.TestHyperH11Integration.test_proxy_connection_close_is_respected`

hyper/common/exceptions.py
Lukasa: Ugh, gross. Well, in that case I think we should just formally rewrite this as taking two arguments: `message`, and `response`.

hyper/http11/response.py
Lukasa: I think we cannot rely on a `proxy-connection: close` header being present. There is no specification requiring it (in fact, `proxy-connection` is not a specified header at all, it's just something that has evolved over time and has been argued as being a "misunderstanding" of how the `connection` header was supposed to work). I don't think there's anything wrong with saying `proxy-connection: close` on a `200 OK` to a `CONNECT`: in fact, if the response is HTTP/1.1 it might even be good practice to provide `Connection: close`, as the termination of the response body is nominally a connection termination.

KostyaEsmukov: So do you suggest to ignore the `proxy-connection` header at all?

How should a response like this one be handled? It causes an assertion error if we ignore the `proxy-connection` header when using an insecure proxy (w/o CONNECT).

    HTTP/1.0 407 Proxy Authentication Required\r\n
    Proxy-Authenticate: Basic realm="proxy"\r\n
    Proxy-Connection: close\r\n
    \r\n

Lukasa: What `AssertionError` does it raise?

KostyaEsmukov: This statement: https://github.com/KostyaEsmukov/hyper/blob/549f5f6d80336d4b76a6532d36ecf2009b67b9be/hyper/http11/response.py#L73-L74

Lukasa: That statement should not be reachable for responses to CONNECT.

KostyaEsmukov: Yes, sure, but this is not a response to CONNECT.

Let me give you a bigger picture on this.
Imagine that we totally ignore the `Proxy-Connection` header.
Consider this situation:

- HTTP/1.1 connection,
- plaintext (`secure=False`),
- with proxy (that is a simple proxying without CONNECT tunnel),
- which requires authentication.

Say we have sent this request to the proxy:

	GET http://stackoverflow.com/ HTTP/1.1
	User-Agent: python-requests/2.13.0
	Accept-Encoding: gzip, deflate
	Accept: */*
	Connection: keep-alive
	proxy-authorization: Basic aGk6aGkK
	connection: Upgrade, HTTP2-Settings
	upgrade: h2c
	HTTP2-Settings: AAQAAP__AAIAAAAA
	host: stackoverflow.com

Proxy might respond with a following:

	HTTP/1.0 407 Proxy Authentication Required
	Proxy-Authenticate: Basic realm="proxy"
	Proxy-Connection: close
	Content-type: text/html; charset=utf-8

	<html><head><title>407 Proxy Authentication Required</title></head>
	<body><h2>407 Proxy Authentication Required</h2><h3>Access to requested resource disallowed by administrator or you need valid username/password to use this resource</h3></body></html>

Note that:
- this is not a response to CONNECT,
- there's no `connection: close` header,
- there's no `transfer-encoding: chunked` header,
- there's no `content-length` header,
- but there's a body, thus I suggest that the `Proxy-Connection: close` header is used akin the `connection: close` header here.

Do you think that there might be a situation when treating the `Proxy-Connection: close` header as `connection: close` would cause unintended effects?

Lukasa: Well, if the proxy responds that way we have a number of problems but none of them matter that much. You see, the connection will be closed by the proxy anyway. In the absence of content-length we expect the connection to be close-delimited, but it *will* be, because that's what `proxy-connection: close` means: at the end of this response I will close the connection.

Further, such a use is largely unprecedented. Proxy-Connection was a mistake, and almost never used. A conforming, remotely-modern proxy will not issue it: instead, it will use the Connection header. This goes *doubly* true in this case when the proxy is sending a response on behalf of itself, rather than on behalf of the remote entity.

I'm inclined to say that as a 21st century HTTP implementation we should simply ignore this header field and behave as though it does not exist.

KostyaEsmukov: But, if we ignore this header, how should we deal with that assertion then?

I believe that such response might violate a standard, but this one I get from a real proxy server, and I'm pretty sure that this is not a single example out there.

So I guess we should work-around such responses somehow. I thought that treating `proxy-connection: close` like a `connection: close` was a decent solution.

Lukasa: I mean, the above response is invalid: Proxy-Connection is not a HTTP/1.1 header that has any effect on framing, so that assertion really should be triggered.

KostyaEsmukov: In the response above a proxy server exclicitly set that this is a HTTP/1.0 response (not HTTP/1.1), didn't set a `Content-Length` header and closed connection after sending body. This is a valid HTTP/1.0 response. 

[RFC 1945, HTTP/1.0, section 7.2.2 Length](https://tools.ietf.org/search/rfc1945#section-7.2.2):
> When an Entity-Body is included with a message, the length of that
> body may be determined in one of two ways. If a Content-Length header
> field is present, its value in bytes represents the length of the
> Entity-Body. Otherwise, the body length is determined by the closing
> of the connection by the server.

[RFC 1945, HTTP/1.0, section 1.3 Overall Operation](https://tools.ietf.org/search/rfc1945#section-1.3):
> Except for experimental applications, current practice requires that
> the connection be established by the client prior to each request and
> closed by the server after sending the response.

But it is an invalid response from HTTP/1.1 perspective as you have already fairly mentioned.

[RFC 7230, HTTP/1.1 Message Syntax and Routing, section 3.4 Handling Incomplete Messages](https://tools.ietf.org/html/rfc7230#section-3.4):

> [...] A response that has neither chunked
> transfer coding nor Content-Length is terminated by closure of the
> connection and, thus, is considered complete regardless of the number
> of message body octets received, provided that the header section was
> received intact.

[RFC 2068, HTTP/1.1, section 14.10 Connection](https://tools.ietf.org/html/rfc2068#section-14.10):
> HTTP/1.1 applications that do not support persistent connections MUST
> include the "close" connection option in every message.
> 
> In HTTP/1.0, each connection is established by the client prior to
> the request and closed by the server after sending the response.
> However, some implementations implement the explicitly negotiated
> ("Keep-Alive") version of persistent connections described in Section
> 19.7.1 of [RFC2068].

Also it is expected, but not strictly required, for HTTP/1.1 clients to understand valid HTTP/1.0 responses as well (as per [RFC 2068, HTTP/1.1, section 19.7 Compatibility with Previous Versions](https://tools.ietf.org/html/rfc2068#section-19.7)).

So the key difference between HTTP/1.1 and HTTP/1.0 responses for that matter, as I see it, is that HTTP/1.1 response must include a `connection: close` header if a server is about to close connection in the absence of `content-length`/chunked encoding headers, but for HTTP/1.0 response connection should always be closed anyway (except when `connection: keep-alive` is present, see below).

I suggest to set `self._expect_close` to True for HTTP/1.0 responses which doesn't contain `connection` header.

[RFC 2068, HTTP/1.1, section 19.7.1 Compatibility with HTTP/1.0 Persistent Connections](https://tools.ietf.org/html/rfc2068#section-19.7.1):

> Some clients and servers may wish to be compatible with some previous
> implementations of persistent connections in HTTP/1.0 clients and
> servers. Persistent connections in HTTP/1.0 must be explicitly
> negotiated as they are not the default behavior. [...]
> 
> The following describes the original HTTP/1.0 form of persistent
> connections.
> 
> When it connects to an origin server, an HTTP client MAY send the
> Keep-Alive connection-token in addition to the Persist connection-
> token:
> 
>       Connection: Keep-Alive
> 
> An HTTP/1.0 server would then respond with the Keep-Alive connection
> token and the client may proceed with an HTTP/1.0 (or Keep-Alive)
> persistent connection.

Thus the assertion will not be triggered even if we ignore the `proxy-conenction` header.

An instant question if you consider this proposal acceptable: should this be done in this PR, or in a different one? This change is not just proxy-related and it definitely requires adding some new tests.

Lukasa: > So the key difference between HTTP/1.1 and HTTP/1.0 responses for that matter, as I see it, is that HTTP/1.1 response must include a connection: close header if a server is about to close connection in the absence of content-length/chunked encoding headers

This is true in the sense that the spec requires it, but the spec also requires that in the absence of all three headers (no `Connection: close`, no `Transfer-Encoding: chunked`, and no `Content-Length`), the response must be treated *as if* `Connection: close` were present. So my suggestion is that we take that route, rather than special-case HTTP/1.0.

KostyaEsmukov: > the spec also requires that in the absence of all three headers (no Connection: close, no Transfer-Encoding: chunked, and no Content-Length), the response must be treated as if Connection: close were present.

I guess this quote confirms what you say:

[RFC 7230, HTTP/1.1 Message Syntax and Routing, section 3.3.3. Message Body Length](https://tools.ietf.org/html/rfc7230#section-3.3.3)
> The length of a message body is determined by one of the following
> (in order of precedence):
> 
> [... cases when either `content-length` or `transfer-encoding: chunked` are set ...]
> 
> 7.  Otherwise, this is a response message without a declared message
>    body length, so the message body length is determined by the
>    number of octets received prior to the server closing the
>    connection.
> 
> Since there is no way to distinguish a successfully completed,
> close-delimited message from a partially received message interrupted
> by network failure, a server SHOULD generate encoding or
> length-delimited messages whenever possible.  The close-delimiting
> feature exists primarily for backwards compatibility with HTTP/1.0.

Here are some extracts that doesn't say the opposite, but it's implicitly meant that in the absence of `Connection: close` header for HTTP/1.1 response a client should persist connection.

[RFC 7230, HTTP/1.1 Message Syntax and Routing, section 6.3. Persistence](https://tools.ietf.org/html/rfc7230#section-6.3):

> o  If the "close" connection option is present, the connection will
>   not persist after the current response; else,
> 
> o  If the received protocol is HTTP/1.1 (or later), the connection
>   will persist after the current response; else,
> 
> o  If the received protocol is HTTP/1.0, the "keep-alive" connection
>   option is present, the recipient is not a proxy, and the recipient
>   wishes to honor the HTTP/1.0 "keep-alive" mechanism, the
>   connection will persist after the current response; otherwise,
> 
> o  The connection will close after the current response.


[RFC 2068, HTTP/1.1, section 8.1.2.1 Negotiation](https://tools.ietf.org/html/rfc2068#section-8.1.2.1):

> An HTTP/1.1 client MAY expect a connection to remain open, but would
> decide to keep it open based on whether the response from a server
> contains a Connection header with the connection-token close. [...]
> 
> If either the client or the server sends the close token in the
> Connection header, that request becomes the last one for the
> connection.
> 
> Clients and servers SHOULD NOT assume that a persistent connection is
> maintained for HTTP versions less than 1.1 unless it is explicitly
> signaled. See section 19.7.1 for more information on backwards
> compatibility with HTTP/1.0 clients.

In sum, we have that in the absence of `content-length` and `transfer-encoding: chunked` headers connection should be closed, but this intent must be explicitly defined with the `connection: close` header. Thus I assume that it's not a valid HTTP/1.1 response and I guess that the assertion we discuss should be triggered for HTTP/1.1 responses, but should not for HTTP/1.0.

Lukasa: Nah, we should tolerate it for HTTP/1.1 as well. 😁

KostyaEsmukov: Okay 😃
But there's a problem: `HTTP/1.0 200 Connection established` response to CONNECT will be considered as close-delimited, so the Response class will eat up the tunnel thinking that this is a body of this response.

I will think about this problem in the coming days. 🙂

Lukasa (changes requested):
Alright, this looks really good! Mind squashing the commits down? When you do that, I'll go ahead and merge.

KostyaEsmukov:
No problem! Should I force-push to this branch, or open a new PR?

I'm asking because I don't know how would GitHub handle the comments in this PR once these commits will be gone (or maybe it is OK to lose them? 😄)

Lukasa:
You can force push to this branch, it'll work just fine.

KostyaEsmukov:
Hi @Lukasa , is anything missing here?

hyper/common/exceptions.py
```diff
@@ -71,3 +71,22 @@ class MissingCertFile(Exception):
...
+
+
+# Create our own ConnectionError.
+try:  # pragma: no cover
+    ConnectionError = ConnectionError
+except NameError:  # pragma: no cover
+    class ConnectionError(Exception):
+        """
+        An error occurred during connection to a host.
+        """
```
Lukasa, line 83: Hrm. Here's a question: on Pythons that have `ConnectionError`, probably we shouldn't create `hyper.common.exceptions.ConnectionError` as an *alias* of that exception but as a subclass instead. Mind having this inherit instead?

Lukasa (changes requested):

hyper/common/exceptions.py
KostyaEsmukov: Why?

Here's the `ConnectionResetError` done the same way in master: https://github.com/Lukasa/hyper/blob/master/hyper/common/exceptions.py#L41

Lukasa: Hrm. Given we have prior art let's keep doing this then. I have mixed feelings about it, but if we're going to fix it we should fix all of it. :grin:

Lukasa:
Ok, if we bring this up-to-date with current master we should be ready to go.

Lukasa (approved):

Lukasa:
Very nicely done! Thanks for spending the time getting this into perfect shape: I know it wasn't easy, and I want you to know I appreciate it greatly.

PR #325, opened 2017-05-29
HTTP/1.1 parser fix, when response contains empty reason

ech0-py:
Sometimes response can come w/o reason-phrase.
Changes made according to the rfc2616 (6.1.1): The client is not required to examine or display the Reason- Phrase.

### DIFF
--- a/hyper/http11/parser.py
+++ b/hyper/http11/parser.py
@@ -49,7 +49,7 @@ class Parser(object):
         if index == -1:
             return None
 
-        version, status, reason = temp_buffer[0:index].split(None, 2)
+        version, status, reason = (temp_buffer[0:index].split(None, 2) + [''])[:3]
         if not version.startswith(b'HTTP/1.'):
             raise ParseError("Not HTTP/1.X!")
 
--- a/hyper/http11/parser.py
+++ b/hyper/http11/parser.py
@@ -49,8 +49,8 @@ class Parser(object):
         if index == -1:
             return None
 
-        version, status, reason = \
-            (temp_buffer[0:index].split(None, 2) + [b''])[:3]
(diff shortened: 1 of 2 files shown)


hyper/http11/parser.py
```diff
@@ -49,7 +49,8 @@ def parse_response(self, buffer):
         if index == -1:
             return None
 
-        version, status, reason = temp_buffer[0:index].split(None, 2)
+        version, status, reason = \
+            (temp_buffer[0:index].split(None, 2) + [b''])[:3]
```
Lukasa: Can we avoid the explicit line continuation here? Let's use Python's implicit line continuation within parentheses and brackets instead.

test/test_http11.py
```diff
@@ -329,6 +329,23 @@ def test_chunked_overrides_body(self):
 
         assert received == expected
 
+    def test_response_with_empty_reason(self):
+        c = HTTP11Connection('httpbin.org')
+        c._sock = sock = DummySocket()
+
+        sock._buffer = BytesIO(
+            b"HTTP/1.1 201\r\n"
```
Lukasa: Interestingly, *this* is not a valid response. You're right that the reason phrase can be absent, but the space in front of it cannot. So this should be `b"HTTP/1.1 201 \r\n"` instead.

Lukasa (changes requested):
Cool, so this is generally a good patch! I've left some notes inline.

ech0-py:
@Lukasa thanks for the advice! Fixed according to the RFC and PEP

Lukasa (approved):
Great, thanks for this! This change looks very reasonable to me.

PR #326, opened 2017-06-06
Check HTTP11 socket is initialized before closing

obrown:
Ensures that the underlying socket exists before calling close on it. This allows us to exit a context safely if connect() hasn't been called

It also matches the behaviour in HTTP20Connection.


Lukasa (approved):
Yup, seems like a good enhancement: thanks!

PR #331, opened 2017-07-02
Fix some chars (!, =, etc) being treated as not allowed in userinfo of authority

KostyaEsmukov:
As per https://tools.ietf.org/html/rfc3986#section-3.2.1

    userinfo    = *( unreserved / pct-encoded / sub-delims / ":" )

Where, according to https://tools.ietf.org/html/rfc3986#section-2.2 ,

    sub-delims  = "!" / "$" / "&" / "'" / "(" / ")"
                      / "*" / "+" / "," / ";" / "="

The `sub-delims` set of chars was added in this PR.

--

(a bit unrelated)

Hyper doesn't handle userinfo to automatically add Basic Auth header to requests. If `host` (or `proxy_host`) of connection contains userinfo, it's getting stripped out in the [to_host_port_tuple function](https://github.com/Lukasa/hyper/blob/e76f185207b4746052b1519cad13238c35195798/hyper/common/util.py#L35).

I feel it's not good to quietly eat up auth part without actually using it. Maybe at least we should issue a warning in this case?

### DIFF
--- a/hyper/packages/rfc3986/misc.py
+++ b/hyper/packages/rfc3986/misc.py
@@ -124,7 +124,7 @@ ip_literal = '\[({0}|{1})\]'.format(ipv6, ipv_future)
 HOST_PATTERN = '({0}|{1}|{2})'.format(reg_name, ipv4, ip_literal)
 
 SUBAUTHORITY_MATCHER = re.compile((
-    '^(?:(?P<userinfo>[A-Za-z0-9_.~\-%:]+)@)?'  # userinfo
+    '^(?:(?P<userinfo>[A-Za-z0-9\-._~%!$&\'()*+,;=:]+)@)?'  # userinfo
     '(?P<host>{0}?)'  # host
     ':?(?P<port>\d+)?$'  # port
     ).format(HOST_PATTERN))
--- a/test/test_hyper.py
+++ b/test/test_hyper.py
@@ -67,6 +67,16 @@ class TestHyperConnection(object):
         assert c.proxy_host == 'localhost'
         assert c.proxy_port == 8443
 
+    def test_connections_can_parse_proxy_hosts_with_userinfo(self):
+        c = HTTP20Connection('www.google.com',
+                             proxy_host='azAz09!==:fakepaswd@localhost:8443')
+        # Note that the userinfo part is getting stripped out,
+        # it's not automatically added as Basic Auth header to
+        # the proxy_headers! It should be done manually.
+        assert c.host == 'www.google.com'
+        assert c.proxy_host == 'localhost'
+        assert c.proxy_port == 8443
+
     def test_connections_can_parse_proxy_hosts_and_ports(self):
         c = HTTP20Connection('www.google.com',
                              proxy_host='localhost',


KostyaEsmukov:
How come first Travis build haven't failed? 😄 Maybe because I've pushed a fix too fast.

That test fails for me locally:



	    def test_connections_can_parse_proxy_hosts_with_userinfo(self):
	        c = HTTP20Connection('www.google.com',
	>                            proxy_host='azAz09!==:fakepaswd@localhost:8443')

	test/test_hyper.py:72: 
	_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
	hyper/http20/connection.py:127: in __init__
	    proxy_host, default_port=8080
	_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

	host_port_str = 'azAz09!==:fakepaswd@localhost:8443', default_port = 8080

	    def to_host_port_tuple(host_port_str, default_port=80):
	        """
	        Converts the given string containing a host and possibly a port
	        to a tuple.
	        """
	        uri = URIReference(
	            scheme=None,
	            authority=host_port_str,
	            path=None,
	            query=None,
	            fragment=None
	        )
	    
	>       host = uri.host.strip('[]')
	E       AttributeError: 'NoneType' object has no attribute 'strip'

	hyper/common/util.py:48: AttributeError

It happens because userinfo part in this test contains `!` and `=` chars, which weren't present in the `SUBAUTHORITY_MATCHER` regex, thus producing None as match

Lukasa:
Thanks for this! However, RFC3986 is a third-party library, not part of this code, so we'd need to fix it in that repo. Sorry!

KostyaEsmukov:
Oh, right, I guess it's this one: https://github.com/sigmavirus24/rfc3986
No problem, I'll open a PR there then.

BTW, what do you think about issuing a warning when stripping out the userinfo part?

Lukasa:
That's the one. 😀

As to the warning, I'm not sure. Try pushing a branch that does it and I'll see what it looks like.

KostyaEsmukov:
This PR has been merged in https://github.com/sigmavirus24/rfc3986/pull/26

What are the following steps? Should I make a PR updating the vendored library?
Or maybe we could wait until this fix is published to the PyPI and unvendor this library? (It was vendored in https://github.com/Lukasa/hyper/pull/172 , but I don't understand why, instead of requiring)

Lukasa:
I'm open to going either direction. The easiest thing to do, probably, is ask for a PyPI release and then unvendor.

KostyaEsmukov:
Great, let's go with unvendoring then. I'll open a PR after PyPI release

PR #335, opened 2017-07-04
Convert headers to strings in HTTP20Adapter to fix #324

### LINKED ISSUE #324: HTTP20Adapter doesn't convert headers from ```bytes``` to ```str```ings
HTTP20Adapter doesn't convert headers and their keys from ```bytes``` to ```str```ings. It breaks compatibility with ```requests``` where headers labels are ```str```ings. Is this by design?

KOLANICH:


hyper/contrib.py
```diff
@@ -132,7 +132,10 @@ def build_response(self, request, resp):
         response = Response()
 
         response.status_code = resp.status
-        response.headers = CaseInsensitiveDict(resp.headers.iter_raw())
+        response.headers = CaseInsensitiveDict((
+            map(lambda h: h.decode("utf-8"), h)
```
Lukasa: This is a little overbroad. To match urllib3 we only want to do this on Python 3: on Python 2 these should stay as bytes.

Lukasa (changes requested):

hyper/contrib.py
```diff
@@ -19,6 +19,10 @@
 from hyper.common.connection import HTTPConnection
 from hyper.compat import urlparse, ssl
 from hyper.tls import init_context
+from sys import version_info as pyver
+
+
+conditionalToStr=lambda s: s.decode("utf-8") if pyver > 2 else lambda s: s
```
Lukasa: Absolutely no need to use a lambda here, let alone two: just use [this function](https://github.com/Lukasa/hyper/blob/development/hyper/common/util.py#L57-L61).

Lukasa (changes requested):

Lukasa (approved):
Looks good to me, thanks! :sparkles:

hyper/contrib.py
```diff
@@ -19,6 +19,7 @@
 from hyper.common.connection import HTTPConnection
 from hyper.compat import urlparse, ssl
 from hyper.tls import init_context
+from hyper.util import to_native_string
```
Lukasa: Oh, this is `hyper.common.util`.

Lukasa (changes requested):

Lukasa (approved):

PR #336, opened 2017-07-04
Added a dictionary of decompressors

KOLANICH:


hyper/http20/response.py
```diff
@@ -71,10 +77,10 @@ def __init__(self, headers, stream):
...
         # Stack Overflow answer for more:
         # http://stackoverflow.com/a/2695466/1401686
-        if b'gzip' in self.headers.get(b'content-encoding', []):
-            self._decompressobj = zlib.decompressobj(16 + zlib.MAX_WBITS)
-        elif b'deflate' in self.headers.get(b'content-encoding', []):
-            self._decompressobj = DeflateDecoder()
+        compressionTypes = set(self.headers.get(b'content-encoding', []))
+        compressionTypes &= set(decompressors.keys())
+        if compressionTypes:
+            self._decompressobj = decompressors[next(iter(compressionTypes))]()
```
Lukasa: There is no reason to do the set operations here. Instead, just use `get` on the dictionary for each value in the `content-encoding` header.

Lukasa (changes requested):

hyper/http20/response.py
```diff
@@ -71,12 +77,12 @@ def __init__(self, headers, stream):
...
-        elif b'deflate' in self.headers.get(b'content-encoding', []):
-            self._decompressobj = DeflateDecoder()
-        else:
-            self._decompressobj = None
+        self._decompressobj = next(
+            (
+                decompressors.get(c)
+                for c in self.headers.get(b'content-encoding', [])
+            )
+        )
```
Lukasa: This doesn't work the way you want it to: if there is no `content-encoding` header, then this will raise `StopIteration`. Rather than try for the clever one-liner, just use a `for` loop. :wink:

Lukasa (changes requested):

hyper/http20/response.py
```diff
@@ -71,12 +77,13 @@ def __init__(self, headers, stream):
...
-        elif b'deflate' in self.headers.get(b'content-encoding', []):
-            self._decompressobj = DeflateDecoder()
-        else:
-            self._decompressobj = None
+        self._decompressobj = next(
+            (
+                decompressors.get(c)
+                for c in self.headers.get(b'content-encoding', [])
+            ),
+            None
```
Lukasa: So at this point I think this is still less clear than just writing a `for` loop with a `break` in it.

Lukasa (changes requested):

hyper/http20/response.py
```diff
@@ -71,12 +77,8 @@ def __init__(self, headers, stream):
...
         # Stack Overflow answer for more:
         # http://stackoverflow.com/a/2695466/1401686
-        if b'gzip' in self.headers.get(b'content-encoding', []):
-            self._decompressobj = zlib.decompressobj(16 + zlib.MAX_WBITS)
-        elif b'deflate' in self.headers.get(b'content-encoding', []):
-            self._decompressobj = DeflateDecoder()
-        else:
-            self._decompressobj = None
+        for c in self.headers.get(b'content-encoding', []):
+            self._decompressobj = decompressors.get(c)
```
Lukasa: Cool, so we should probably also add an early exit here if we have a non-`None` value.

Lukasa (changes requested):

hyper/http20/response.py
Lukasa: Nope, this is overbroad: we now `break` after the *first* attempt to get a decompressor, even if the content-encoding used is not a compression one. We need to check that `self._decompressobj is not None` before breaking.

Lukasa (changes requested):

Lukasa (approved):

hyper/http20/response.py
```diff
@@ -71,12 +77,9 @@ def __init__(self, headers, stream):
         # This 16 + MAX_WBITS nonsense is to force gzip. See this
         # Stack Overflow answer for more:
         # http://stackoverflow.com/a/2695466/1401686
-        if b'gzip' in self.headers.get(b'content-encoding', []):
-            self._decompressobj = zlib.decompressobj(16 + zlib.MAX_WBITS)
-        elif b'deflate' in self.headers.get(b'content-encoding', []):
-            self._decompressobj = DeflateDecoder()
-        else:
-            self._decompressobj = None
+        for c in self.headers.get(b'content-encoding', []):
```
Lukasa, line 81: If this list is empty the `_decompressobj` never gets assigned. It's sufficient to set it to `None` at class scope to resolve this.

Lukasa (changes requested):

Lukasa (approved):
Ok, this looks reasonable to me.

PR #337, opened 2017-07-04
Brotli

KOLANICH:
Added brotli decompression

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -88,7 +88,7 @@ setup(
     },
     extras_require={
         'fast': ['pycohttpparser'],
-        'brotli decompression': ['brotlipy'],
+        'brotli': ['brotlipy'],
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -88,6 +88,7 @@ def run_tests(self):
     },
     extras_require={
         'fast': ['pycohttpparser'],
+        'brotli decompression': ['brotlipy'],
```
Lukasa: I'd just call this `brotli`.

Lukasa (changes requested):

Lukasa:
Ok, so this is fine, contingent on being rebased on top of #336 when that is finished.

Lukasa:
Can you rebase this change onto the current `development` please?

PR #339, opened 2017-07-05
Brotli

KOLANICH:


Lukasa (changes requested):
Looks like we need some tests that actually exercise the brotli functionality.

KOLANICH:
Yes, but I'm not going to write them.

Lukasa:
Ok, this will need to sit here until someone gets around to writing tests for it. 😁 Thanks for the work!

peterbe:
Can we close this now? 
Seems it's already there: https://github.com/Lukasa/hyper/blob/18b629b8487169870235fe13387e2ae03c178b9f/setup.py#L80 and https://github.com/Lukasa/hyper/blob/18b629b8487169870235fe13387e2ae03c178b9f/hyper/http20/response.py#L35

PR #341, opened 2017-07-21
Unvendor rfc3986

KostyaEsmukov:
Unvendor rfc3986 package. This also fixes an issue with some symbols treated as invalid parts of userinfo of authority when they are valid (the fix is in https://github.com/sigmavirus24/rfc3986/pull/26)

Follow up for #331


sigmavirus24 (approved):

Lukasa (approved):
Very nice! This looks great to me. :sparkles:

PR #342, opened 2017-07-24
Add connection/read timeout for requests adapter

hyxbiao:
[+] Add connection/read timeout for requests adapter, we can use like this:

```
import requests
from hyper.contrib import HTTP20Adapter
s = requests.Session()
s.mount('https://http2bin.org', HTTP20Adapter())
r = s.get('https://http2bin.org/get', timeout=30)
print(r.status_code)
```

### DIFF
--- a/hyper/common/connection.py
+++ b/hyper/common/connection.py
@@ -58,6 +58,7 @@ class HTTPConnection(object):
                  proxy_host=None,
                  proxy_port=None,
                  proxy_headers=None,
+                 timeout=None,
                  **kwargs):
 
         self._host = host
@@ -78,8 +79,10 @@ class HTTPConnection(object):
         self._h1_kwargs.update(kwargs)
         self._h2_kwargs.update(kwargs)
 
+        self._timeout = timeout
+
         self._conn = HTTP11Connection(
-            self._host, self._port, **self._h1_kwargs
+            self._host, self._port, timeout=self._timeout, **self._h1_kwargs
         )
 
     def request(self, method, url, body=None, headers=None):
@@ -113,7 +116,8 @@ class HTTPConnection(object):
             assert e.negotiated in H2_NPN_PROTOCOLS
 
             self._conn = HTTP20Connection(
-                self._host, self._port, **self._h2_kwargs
+                self._host, self._port,
+                timeout=self._timeout, **self._h2_kwargs
             )
             self._conn._sock = e.sock
 
@@ -138,7 +142,8 @@ class HTTPConnection(object):
             assert e.negotiated == H2C_PROTOCOL
 
             self._conn = HTTP20Connection(
-                self._host, self._port, **self._h2_kwargs
+                self._host, self._port,
+                timeout=self._timeout, **self._h2_kwargs
             )
 
             self._conn._connect_upgrade(e.sock)
--- a/hyper/common/connection.py
+++ b/hyper/common/connection.py
@@ -66,23 +66,23 @@ class HTTPConnection(object):
         self._h1_kwargs = {
             'secure': secure, 'ssl_context': ssl_context,
             'proxy_host': proxy_host, 'proxy_port': proxy_port,
-            'proxy_headers': proxy_headers, 'enable_push': enable_push
+            'proxy_headers': proxy_headers, 'enable_push': enable_push,
+            'timeout': timeout
         }
         self._h2_kwargs = {
             'window_manager': window_manager, 'enable_push': enable_push,
             'secure': secure, 'ssl_context': ssl_context,
             'proxy_host': proxy_host, 'proxy_port': proxy_port,
-            'proxy_headers': proxy_headers
+            'proxy_headers': proxy_headers,
+            'timeout': timeout
         }
 
         # Add any unexpected kwargs to both dictionaries.
         self._h1_kwargs.update(kwargs)
         self._h2_kwargs.update(kwargs)
 
-        self._timeout = timeout
-
         self._conn = HTTP11Connection(
-            self._host, self._port, timeout=self._timeout, **self._h1_kwargs
+            self._host, self._port, **self._h1_kwargs
         )
 
     def request(self, method, url, body=None, headers=None):
@@ -116,8 +116,7 @@ class HTTPConnection(object):
             assert e.negotiated in H2_NPN_PROTOCOLS
 
             self._conn = HTTP20Connection(
-                self._host, self._port,
-                timeout=self._timeout, **self._h2_kwargs
+                self._host, self._port, **self._h2_kwargs
             )
             self._conn._sock = e.sock
 
@@ -142,8 +141,7 @@ class HTTPConnection(object):
             assert e.negotiated == H2C_PROTOCOL
 
             self._conn = HTTP20Connection(
-                self._host, self._port,
-                timeout=self._timeout, **self._h2_kwargs
+                self._host, self._port, **self._h2_kwargs
             )
 
             self._conn._connect_upgrade(e.sock)
--- a/hyper/contrib.py
+++ b/hyper/contrib.py
@@ -33,7 +33,7 @@ class HTTP20Adapter(HTTPAdapter):
         self.connections = {}
 
     def get_connection(self, host, port, scheme, cert=None, verify=True,
-                       proxy=None):
+                       proxy=None, **kwargs):
         """
         Gets an appropriate HTTP/2 connection object based on
         host/port/scheme/cert tuples.
@@ -77,7 +77,8 @@ class HTTP20Adapter(HTTPAdapter):
                 secure=secure,
                 ssl_context=ssl_context,
                 proxy_host=proxy_netloc,
-                proxy_headers=proxy_headers)
+                proxy_headers=proxy_headers,
+                **kwargs)
             self.connections[connection_key] = conn
 
         return conn
@@ -98,7 +99,8 @@ class HTTP20Adapter(HTTPAdapter):
             parsed.scheme,
             cert=cert,
             verify=verify,
-            proxy=proxy)
+            proxy=proxy,
+            **kwargs)
 
         # Build the selector.
         selector = parsed.path
--- a/hyper/contrib.py
+++ b/hyper/contrib.py
@@ -33,7 +33,7 @@ class HTTP20Adapter(HTTPAdapter):
         self.connections = {}
 
     def get_connection(self, host, port, scheme, cert=None, verify=True,
-                       proxy=None, **kwargs):
+                       proxy=None, timeout=None):
         """
         Gets an appropriate HTTP/2 connection object based on
         host/port/scheme/cert tuples.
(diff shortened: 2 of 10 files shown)


hyper/contrib.py
```diff
@@ -77,7 +77,8 @@ def get_connection(self, host, port, scheme, cert=None, verify=True,
                 secure=secure,
                 ssl_context=ssl_context,
                 proxy_host=proxy_netloc,
-                proxy_headers=proxy_headers)
+                proxy_headers=proxy_headers,
+                **kwargs)
```
Lukasa: We shouldn't be sending arbitrary kwargs from Requests to hyper: there's no reason to assume that those kwargs will match in any sensible way. We should aim to be explicit here.

Lukasa (changes requested):
Thanks for this! I've left a note inline. It'd also be really useful to actually add tests for the timeout functionality that we're implicitly also adding to our lower-level classes.

hyper/http11/connection.py
```diff
@@ -150,6 +150,15 @@ def __init__(self, host, port=None, secure=None, ssl_context=None,
         #: the standard hyper parsing interface.
         self.parser = Parser()
 
+        # timeout
+        timeout = kwargs.get('timeout')
```
KostyaEsmukov: `timeout` should be optional. I think it should be `None` - that means no timeout.

KostyaEsmukov (changes requested):
Thanks for this! I was going to work on timeouts this week but you have started earlier :)

I have one more note on this: the `_create_tunnel` function of http11 module should respect the `timeout` too.

PS: Related issue: #187

hyxbiao:
timeout is optional, the default value is None, means that it will block connection/read, it is the same as request/adapters.py:
`def send(self, request, stream=False, timeout=None, verify=True, cert=None, proxies=None):`
```
if isinstance(timeout, tuple):
    try:
        connect, read = timeout
        timeout = TimeoutSauce(connect=connect, read=read)
    except ValueError as e:
        # this may raise a string formatting error.
        err = ("Invalid timeout {0}. Pass a (connect, read) "
               "timeout tuple, or a single float to set "
               "both timeouts to the same value".format(timeout))
        raise ValueError(err)
else:
    timeout = TimeoutSauce(connect=timeout, read=timeout)
```

it also can set timeout to socket._GLOBAL_DEFAULT_TIMEOUT

hyper/contrib.py
hyxbiao: OK, I had change to explicit timeout and add more timeout tests

hyper/http11/connection.py
hyxbiao: Yes, timeout is optional and default None, is equivalent to socket.setblocking(1)

hyper/contrib.py
```diff
@@ -92,13 +93,15 @@ def send(self, request, stream=False, cert=None, verify=True, proxies=None,
             proxy = prepend_scheme_if_needed(proxy, 'http')
 
         parsed = urlparse(request.url)
+        timeout = kwargs.get('timeout')
```
Lukasa: Let's just add `timeout` to our list of kwargs.

test/test_integration.py
```diff
@@ -1230,6 +1231,114 @@ def do_connect(conn):
...
+        def socket_handler(listener):
+            time.sleep(1)
+            sock = listener.accept()[0]
+            sock.close()
+
+        self._start_server(socket_handler)
+        conn = self.get_connection()
+        try:
+            conn.connect()
+            assert False
```
Lukasa: This should be `pytest.fail`, and generally should go in an `else` block.

test/test_integration.py
```diff
@@ -1230,6 +1231,114 @@ def do_connect(conn):
...
+            sock = listener.accept()[0]
+
+            # We get two messages for the connection open and then a HEADERS
+            # frame.
+            receive_preamble(sock)
+            sock.recv(65535)
+
+            # Wait for request
+            req_event.wait(5)
+            # Now, send the headers for the response. This response has no body
```
Lukasa: This comment is misleading.

test/test_integration.py
```diff
@@ -1230,6 +1231,114 @@ def do_connect(conn):
...
+            time.sleep(1)
+
+            # Now, send the headers for the response. This response has no body
+            f = build_headers_frame(
```
Lukasa: We never actually need to send this, and doing so could cause exceptions, so let's not send any data here.

test/test_integration.py
```diff
@@ -1230,6 +1231,114 @@ def do_connect(conn):
...
+            sock.close()
+
+        self._start_server(socket_handler)
+        conn = self.get_connection()
+        conn.request('GET', '/')
+        req_event.set()
+
+        try:
+            conn.get_response()
+            assert False
```
Lukasa: Same note about `pytest.fail` and `else` blocks.

test/test_integration.py
```diff
@@ -1230,6 +1231,114 @@ def do_connect(conn):
...
+        try:
+            conn.connect()
+        except (SocketTimeout, ssl.SSLError):
+            # Py2 raises this as a BaseSSLError,
+            # Py3 raises it as socket timeout.
+            assert False
+
+        send_event.wait(5)
+
+        assert data[0].startswith(b'PRI * HTTP/2.0\r\n\r\nSM\r\n\r\n')
```
Lukasa: How can this possibly pass? If the connection fails, we're never going to see this data.

test/test_integration.py
```diff
@@ -1230,6 +1231,114 @@ def do_connect(conn):
 
         self.tear_down()
 
+    def test_connection_timeout(self):
+        self.set_up(timeout=0.5)
+
+        def socket_handler(listener):
+            time.sleep(1)
+            sock = listener.accept()[0]
```
Lukasa: No point calling `accept`: it'll usually fail, and throw exceptions, which we don't want.

Lukasa (changes requested):

test/test_integration.py
hyxbiao: default timeout `None` will block connect, so it will connect sucess and go on

test/test_integration.py
hyxbiao: if connection timeout smaller than 1s, it will throw timeout exception, in this test connection timeout is set to 0.5, so it will throw exception as expect

hyper/http11/connection.py
```diff
@@ -101,7 +101,7 @@ class HTTP11Connection(object):
 
     def __init__(self, host, port=None, secure=None, ssl_context=None,
                  proxy_host=None, proxy_port=None, proxy_headers=None,
-                 **kwargs):
+                 timeout=None, **kwargs):
```
Lukasa, line 104: We should probably add this to the common `HTTPConnection` object as well.

test/test_http11.py
```diff
@@ -110,6 +110,18 @@ def test_initialization_with_ipv6_addresses_proxy_inline_port(self):
...
+
+        assert c._connect_timeout == 30
+        assert c._read_timeout == 30
+
+    def test_initialization_tuple_timeout(self):
+        c = HTTP11Connection('httpbin.org', timeout=(5, 60))
+
+        assert c._connect_timeout == 5
+        assert c._read_timeout == 60
+
```
Lukasa: We need these tests for HTTP/2 as well.

test/test_integration.py
Lukasa: Right, but this exception is not caught, meaning it'll be logged and generally treated badly. We shouldn't do anything in the background thread that we *know* will fail, and `accept` *will* fail here.

test/test_integration.py
```diff
@@ -1230,6 +1231,104 @@ def do_connect(conn):
...
+        conn = self.get_connection()
+        try:
+            conn.connect()
+        except (SocketTimeout, ssl.SSLError):
+            # Py2 raises this as a BaseSSLError,
+            # Py3 raises it as socket timeout.
+            # assert 'timed out' in e.message
+            pass
+        else:
+            assert False
```
Lukasa: `pytest.fail`, please.

test/test_integration.py
```diff
@@ -1230,6 +1231,104 @@ def do_connect(conn):
...
+            sock.close()
+
+        self._start_server(socket_handler)
+        conn = self.get_connection()
+        try:
+            conn.connect()
+        except (SocketTimeout, ssl.SSLError):
+            # Py2 raises this as a BaseSSLError,
+            # Py3 raises it as socket timeout.
+            # assert 'timed out' in e.message
```
Lukasa: Probably we don't need the `assert` statement in the comment?

test/test_integration.py
```diff
@@ -1230,6 +1231,104 @@ def do_connect(conn):
...
+
+        try:
+            conn.get_response()
+        except (SocketTimeout, ssl.SSLError):
+            # Py2 raises this as a BaseSSLError,
+            # Py3 raises it as socket timeout.
+            # assert 'timed out' in e.message
+            pass
+        else:
+            assert False
```
Lukasa: `pytest.fail`, please.

test/test_integration.py
```diff
@@ -1230,6 +1231,104 @@ def do_connect(conn):
...
+            sock.close()
+
+        self._start_server(socket_handler)
+        conn = self.get_connection()
+        try:
+            conn.connect()
+        except (SocketTimeout, ssl.SSLError):
+            # Py2 raises this as a BaseSSLError,
+            # Py3 raises it as socket timeout.
+            assert False
```
Lukasa: `pytest.fail`, please.

test/test_integration.py
Lukasa: Ah, sorry, I see that. Thanks!

Lukasa (changes requested):

test/test_integration.py
hyxbiao: sorry, i'm not familiar with pytest before, i replace it with pytest.raises or pytest.fail

hyper/http11/connection.py
hyxbiao: ok，I just added

test/test_http11.py
hyxbiao: ok，I add this in test/test_hyper.py

test/test_integration.py
hyxbiao: got it, I just updated

test/test_integration.py
Lukasa: `pytest.raises` is the best thing to do if you don't plan to do anything in the `except` block.

hyper/common/connection.py
```diff
@@ -78,8 +79,10 @@ def __init__(self,
         self._h1_kwargs.update(kwargs)
         self._h2_kwargs.update(kwargs)
 
+        self._timeout = timeout
```
Lukasa: This should go into `_h1_kwargs` and `_h2_kwargs`: it'll also need associated tests.

Lukasa (changes requested):

hyxbiao:
update information:
add timeout for `_create_tunnel` funciton, almost forget it
move timeout to _h1/2_kwargs in hyper/common/connection.py
add more timeout tests

hyper/common/connection.py
hyxbiao: ok, it's done

Lukasa (approved):
Very nicely done, thank you!

PR #344, opened 2017-07-31
HTTP20Adapter.close tests (supersedes #307)

KostyaEsmukov:
Add `close()` method implementation to the requests adapter and integration tests for it.

Supersedes #307
Related issue: #306

### DIFF
--- a/test/test_integration.py
+++ b/test/test_integration.py
@@ -1742,10 +1742,19 @@ class TestRequestsAdapter(SocketLevelTest):
 
         self._start_server(socket_handler)
 
+        a = HTTP20Adapter()
         s = requests.Session()
-        s.mount('http://', HTTP20Adapter())
+        s.mount('http://', a)
         r = s.get('http://%s:%s' % (self.host, self.port))
+        connections_before_close = list(a.connections.values())
         s.close()
+
+        # check that connections cache is empty
+        assert not a.connections
+
+        # check that all connections are actually closed
+        assert (connections_before_close and
+                all(conn._sock is None for conn in connections_before_close))
 
         assert r.status_code == 201
         assert len(r.headers) == 3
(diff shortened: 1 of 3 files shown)


Lukasa (changes requested):
Thanks for this! It would be good if we could add a test that demonstrates that calling `close` on the adapter both closes the underlying connections and also clears the connection cache. Mind adding those?

KostyaEsmukov:
I've added that check to these two tests instead of creating a new one. Let me know if you think that this should be checked in a separate test.

test/test_integration.py
```diff
@@ -1717,3 +1717,98 @@ def socket_handler(listener):
...
+        r = s.get('http://%s:%s' % (self.host, self.port))
+        connections_before_close = list(a.connections.values())
+        s.close()
+
+        # check that connections cache is empty
+        assert not a.connections
+
+        # check that all connections are actually closed
+        assert (connections_before_close and
+                all(conn._sock is None for conn in connections_before_close))
```
Lukasa: Let's split the first assertion out and call it when we get the connections initially.

test/test_integration.py
```diff
@@ -1717,3 +1717,98 @@ def socket_handler(listener):
...
+            s.mount('http://', a)
+            r = s.get('http://%s:%s' % (self.host, self.port))
+            connections_before_close = list(a.connections.values())
+
+        # check that connections cache is empty
+        assert not a.connections
+
+        # check that all connections are actually closed
+        assert (connections_before_close and
+                all(conn._sock is None for conn in connections_before_close))
```
Lukasa: Same note here.

Lukasa (changes requested):

Lukasa (approved):
Very nice, this looks really good. Thanks so much!

PR #347, opened 2017-08-30
Fix connection upgrade to work for both python2 and python3

ericdevries:
I had an issue with this line of code not working in a python3 package, because the code in the h2 package differs between these 2 versions of python

### DIFF
--- a/hyper/http20/connection.py
+++ b/hyper/http20/connection.py
@@ -403,7 +403,7 @@ class HTTP20Connection(object):
         with self._conn as conn:
             conn.initiate_upgrade_connection()
             conn.update_settings(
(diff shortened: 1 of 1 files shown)


Lukasa:
@ericdevries I think you mis-identified the problem here. I think either your Python 3 or Python 2 environments are running an unsupported copy of `h2`: this code does not support `h2` 3.0 or higher, which is where this change would be needed.

ericdevries:
What do you mean? h2 has to fix it or your code has to be made compatible with h2 version 3.0+?

Lukasa:
Eventually hyper needs to be made compatible with h2 version 3.0, but for the moment our declared dependency versions exclude 3.0.

Issue #355, opened 2017-10-26
END_STREAM not marked when last chunk of length MAX_CHUNK


PrimozGodec:
https://github.com/Lukasa/hyper/blob/13e2a909e14349d14ae2b86217062e1568242b9a/hyper/http20/stream.py#L330

End of the stream is not marked when the last chunk is of size equal to MAX_CHUNK (1024). Condition `len(data) < MAX_CHUNK` works fine in all cases when the last chunk is of size smaller than 1024 but not in case when last chunk is of size 1024. If we change this to `len(data) <= MAX_CHUNK` it wont work since we will mark END_STREAM for every chunk. Suggested solution is to make another argument that tell you when you have a last chunk.

PR #356, opened 2017-10-26
Fix: stream end when length of last chunk equal to MAX_CHUNK

PrimozGodec:
This pull request is fixing bug https://github.com/Lukasa/hyper/issues/355

I changed the approach of taking chunks form the generator, such that we know when we reach the last chunk with memorising the next chunk.

### DIFF
--- a/hyper/http20/stream.py
+++ b/hyper/http20/stream.py
@@ -122,8 +122,18 @@ class Stream(object):
             chunks = (data[i:i+MAX_CHUNK]
                       for i in range(0, len(data), MAX_CHUNK))
 
-        for chunk in chunks:
-            self._send_chunk(chunk, final)
+        # since we need to know when we have a last package we need to know
+        # if there is another package in advance
+        cur_chunk = None
+        try:
+            cur_chunk = next(chunks)
+            while True:
+                next_chunk = next(chunks)
+                self._send_chunk(cur_chunk, False)
+                cur_chunk = next_chunk
+        except StopIteration:
+            if cur_chunk is not None:  # cur_chunk none when no chunks to send
+                self._send_chunk(cur_chunk, final)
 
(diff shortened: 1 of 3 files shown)


Lukasa (changes requested):
Yup, this looks like it’ll work. Mind adding an entry to the change log and also a test for me?

PrimozGodec:
@Lukasa I changed the code and made unit test pass on my machine. I also added a test to test specific change. 
Sorry I can not find a change log. Can you tell me where is it?
Also travis tests are failing it seems that they are failing because of some older code and not because of my changes.

matjazp:
@PrimozGodec, CI is fixed now, can you rebase please?

Also, change log is in HISTORY.rst and CONTRIBUTORS.rst.

PrimozGodec:
Thank you @matjazp. It is fixed now @Lukasa

hyper/http20/stream.py
```diff
@@ -122,8 +122,17 @@ def file_iterator(fobj):
             chunks = (data[i:i+MAX_CHUNK]
                       for i in range(0, len(data), MAX_CHUNK))
 
-        for chunk in chunks:
-            self._send_chunk(chunk, final)
+        # since we need to know when we have a last package we need to know
+        # if there is another package in advance
+        cur_chunk = next(chunks)
```
Lukasa: This will explode if there are no chunks to send: I think you need to write a test case for that and refactor.

Lukasa (changes requested):

hyper/http20/stream.py
PrimozGodec: Done

Lukasa (approved):
Looks good to me, thanks! :sparkles:

PR #361, opened 2017-11-04
fix zlib_compressobj wbit

johejo:
24 is not appropriate for the value of WBITS
https://docs.python.jp/3/library/zlib.html#zlib.compressobj


Lukasa (approved):
Thanks!

PR #366, opened 2017-11-11
Fix http11 head response

johejo:
- Dealing with the problem of reading an empty body from HTTP / 1.1 HEAD response.
- Add test function related to the problem.


Lukasa (approved):
Looks good to me, thanks so much!

PR #370, opened 2017-12-12
Add a fix to make it possible to add window_manager to HTTP20Adapter.

johejo:
I wanted to use a custom window_manager for HTTP20Adapter.
However, I feel that this change is not very appropriate.
Please let us know if you have any advice.


Lukasa:
I think there's no problem with this at all. Thanks! :sparkles:

PR #386, opened 2018-06-14
Add '@=' separator to pass JSON objects via CLI

agoconcept:
The current '=' separator enables to add key-value pairs as strings
in a JSON object to the HTTP body. But there's no way to provide
more complex JSON objects

This commit introduces a new '@=' that adds a key-value pair where
the right value is evaluated to a JSON object, allowing more complex
structures in the HTTP body.

Change-Id: I126a6c8529a6dbf9108b904b2b4d93ef60309113

### DIFF
--- a/hyper/cli.py
+++ b/hyper/cli.py
@@ -101,7 +101,7 @@ def make_positional_argument(parser):
 
         '@=' JSON data fields
 
-            name='{"name": "John", "surname": "Doe"}' list='[1, 2, 3]'
+            name@='{"name": "John", "surname": "Doe"}' list@='[1, 2, 3]'
         """))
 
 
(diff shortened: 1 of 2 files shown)


hyper/cli.py
```diff
@@ -93,9 +95,13 @@ def make_positional_argument(parser):
...
             search==hyper
 
-        '=' Data fields to be serialized into a JSON object:
+        '=' String data fields to be serialized into a JSON object:
 
             name=Hyper  language=Python  description='CLI HTTP client'
+
+        '@=' JSON data fields
+
+            name='{"name": "John", "surname": "Doe"}' list='[1, 2, 3]'
```
KostyaEsmukov: Shouldn't it have `@=` there too, like `name@='{"name": "John", "surname": "Doe"}' list@='[1, 2, 3]'`, or am I missing something?

agoconcept: You're completely right, I'll fix that. Thanks for noticing

agoconcept:
Is there any additional change needed to accept the pull request? In case anything else is needed please don't hesitate to ask for it

Cheers!

PR #405, opened 2019-04-12
python3 compatibility fix for FakeOriginalResponse

xingzhisg:
The FakeOriginalResponse needs a fix for python3 compatibility


divinehawk:
Please merge. This also fixes cookie parsing when using both Python 2 and 3.

xingzhisg:
Indeed this patch was for fixing the cookie setting issue in python3. Sorry I didn't highlight the keyword 'cookie' in the original post.

PR #429, opened 2020-03-19
Support for latest h2 version

zetahernandez:


dashdanw:
@zetahernandez maybe re-run ci?

